PR #73, opened 2026-09-03
fix(033): assert the V-014 DFS invariants instead of silently falling back

bartekus:
Follow-up to #72, from the AI review on that PR. Hardening only: no release
behavior changes.

`detect_dependency_cycle` has two fallbacks that are unreachable today and
would fail *quietly* rather than loudly if that ever stopped being true.

**The GREY arm.** `stack.iter().position(...).unwrap_or(0)` rests on the
invariant "GREY exactly while a node is on the current stack" (set on push,
cleared to BLACK on pop). If that broke, `from = 0` would still report a real
cycle, but would name the entire stack and point the violation at the DFS root
instead of the spec the cycle re-enters: a plausible-looking wrong path.

**The outer loop.** `color.get(...).unwrap_or(BLACK)` rests on `color` being
seeded from every record id. If that broke, the start node would be skipped
and its cycle missed entirely.

## Why `debug_assert!` and not `expect`

The review suggested `expect("V-014: GREY node not on stack")`. Same goal,
wrong instrument for this crate:

- Core is documented panic-free on user input (`CLAUDE.md`, and the
  constitution's determinism/validation principle). The only two `.expect()`
  calls in core production code, `query.rs:32` and `shard.rs:33`, both assert
  on a **compile-time constant**, never on runtime-derived state. An `expect`
  here would be the first panic in core gated on graph shape.
- `spec-spine` is a CI gate. Aborting the run with a backtrace is a worse
  outcome than the fallback, which still reports a genuine cycle (reaching the
  GREY arm at all means `child` is on the current path) and only misnames the
  path.

`debug_assert!` gets the review's actual intent without either cost: it is live
in every test and debug build, so a future refactor that breaks an invariant
fails `cargo test` loudly, and release keeps the graceful degradation. All 40
compile tests pass with both assertions active.

## Review findings not acted on

| finding | disposition |
|---|---|
| Self-dependency: increment before cycle detection is dead on the returning path | No change. The review itself calls it "not a bug". The increment is load-bearing for the WHITE and BLACK arms; hoisting it would complicate the loop for no gain. `X -> X` already reports correctly. |
| `016-short-id-resolution` asserted but not evidenced in the diff | False positive. It exists and is `approved` (`registry show 016-short-id-resolution`), and `lint --fail-on-warn --fail-on-info` is 0/0/0, so `V-010` does not fire. The review saw the diff, not the corpus. |
| `BTreeMap`/`BTreeSet` costs `O(log n)` where a hash map is `O(1)` | No change, as the review's own conclusion allows. On a 34-spec corpus the difference is unmeasurable, and `BTreeMap` is the file's uniform style in a crate whose central claim is deterministic output. Swapping in `HashMap` here is safe only because none of these maps is currently iterated; keeping the ordered type removes that as a trap for a future edit that does iterate. |

## Verification

| check | result |
|---|---|
| `compile --check` | fresh, 34 shards |
| `index check` | fresh (confirmed: regenerating both artifacts produces no diff) |
| `couple --base main --head HEAD` | 1 violation, waived below |
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test --workspace --locked` | all pass |

Spec-Drift-Waiver: crates/spec-spine-core/src/compile.rs - adds two `debug_assert!` guards over already-unreachable fallbacks in `detect_dependency_cycle`. No release behavior changes and no claim in any of the nine owning specs (001, 013, 014, 016, 018, 019, 024, 031, 033) is affected, so amending one would be a mechanical refresh, which `.claude/rules/adversarial-prompt-refusal.md` explicitly forbids.

### DIFF
--- a/crates/spec-spine-core/src/compile.rs
+++ b/crates/spec-spine-core/src/compile.rs
@@ -439,28 +439,34 @@ fn detect_dependency_cycle(records: &[SpecRecord], out: &mut Vec<Violation>) {
                     // `child` is still on the current path, so everything from
                     // its position up to the top of the stack is the cycle;
                     // naming it once more closes the path into a loop.
-                    // GREY holds exactly while a node is on the current
-                    // stack: set on push, cleared to BLACK on pop. The search
-                    // therefore always hits. Assert it in debug, because the
-                    // fallback would still report a real cycle but would name
-                    // the DFS root instead of the spec the cycle re-enters.
+                    //
+                    // GREY holds exactly while a node is on the current stack:
+                    // set on push, cleared to BLACK on pop. The search
+                    // therefore always hits, and `debug_assert` makes a broken
+                    // invariant fail `cargo test` loudly. Release does not
+                    // panic (that would abort the gate outside its documented
+                    // exit codes) and does not invent a path either: reaching
+                    // this arm proves a cycle through `child` exists, so the
+                    // fallback reports that much and drops only the precision
+                    // it can no longer vouch for.
                     let entry = stack.iter().position(|&(n, _)| n == child);
                     debug_assert!(
                         entry.is_some(),
                         "V-014: GREY node absent from the current stack"
                     );
-                    let from = entry.unwrap_or(0);
-                    let mut cycle: Vec<&str> = stack[from..].iter().map(|&(n, _)| n).collect();
-                    cycle.push(child);
-                    let at_path = cycle
-                        .first()
-                        .and_then(|id| spec_paths.get(*id))
-                        .map(|p| p.to_string());
-                    out.push(error(
-                        "V-014",
-                        format!("depends_on cycle: {}", cycle.join(" -> ")),
-                        at_path,
-                    ));
+                    // `child` is the spec the cycle re-enters, so it is on the
+                    // cycle whichever branch below runs.
+                    let at_path = spec_paths.get(child).map(|p| p.to_string());
+                    let message = match entry {
+                        Some(from) => {
+                            let mut cycle: Vec<&str> =
+                                stack[from..].iter().map(|&(n, _)| n).collect();
+                            cycle.push(child);
+                            format!("depends_on cycle: {}", cycle.join(" -> "))
+                        }
+                        None => format!("depends_on cycle through '{child}'"),
+                    };
+                    out.push(error("V-014", message, at_path));
                     return;
                 }
                 _ => {}
--- a/crates/spec-spine-core/src/compile.rs
+++ b/crates/spec-spine-core/src/compile.rs
@@ -408,14 +408,19 @@ fn detect_dependency_cycle(records: &[SpecRecord], out: &mut Vec<Violation>) {
     let mut color: BTreeMap<&str, u8> = ids.iter().map(|id| (*id, WHITE)).collect();
 
     for r in records {
-        // Every record id was seeded into `color` above, so the fallback is
-        // unreachable. Assert it in debug rather than let a future change
-        // silently skip a start node and miss a cycle entirely.
+        // `color` is seeded from `ids`, which is collected from these same
+        // `records` four lines up, and no later insert adds a key (every
+        // `child` is filtered through `ids`). The fallback is therefore
+        // unreachable. It defaults to WHITE rather than BLACK so that if it
+        // ever became reachable the node would still be walked: a redundant
+        // walk is harmless, whereas defaulting to BLACK would skip the node
+        // and miss a cycle rooted there, which is the one outcome a gate must
+        // never produce silently.
         debug_assert!(
             color.contains_key(r.id.as_str()),
-            "V-014: record id absent from the color map"
+            "dependency-cycle detector: record id absent from the color map"
         );
-        if color.get(r.id.as_str()).copied().unwrap_or(BLACK) != WHITE {
+        if color.get(r.id.as_str()).copied().unwrap_or(WHITE) != WHITE {
             continue;
         }
         // (spec id, index of the next depends_on entry to walk from it).
@@ -452,7 +457,7 @@ fn detect_dependency_cycle(records: &[SpecRecord], out: &mut Vec<Violation>) {
                     let entry = stack.iter().position(|&(n, _)| n == child);
                     debug_assert!(
                         entry.is_some(),
-                        "V-014: GREY node absent from the current stack"
+                        "dependency-cycle detector: GREY node absent from the stack"
                     );
                     // `child` is the spec the cycle re-enters, so it is on the
                     // cycle whichever branch below runs.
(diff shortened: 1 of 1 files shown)


bartekus:
Thanks: one of the four points landed, and I have pushed a fix for it. The other three are artifacts of reviewing the 19-line diff without the surrounding function, the PR body, or the CI result.

## Acted on: the release-build silence

You are right that `debug_assert!` is stripped in release, so a broken invariant would let release report a wrong cycle. The answer is to remove the wrongness, not to trade it for a panic.

The GREY arm now reports only what reaching it proves. Entering that arm means `child` is on the current path, so a cycle through `child` exists whether or not the stack search hits:

- search hits (always, today): the full path, exactly as before.
- search misses: `depends_on cycle through '<child>'`, dropping only the precision the code can no longer vouch for, instead of splicing `stack[0..]` and presenting an invented path as fact.

`at_path` now comes from `child` directly. On the reachable path that is identical (`cycle.first()` is `child` by construction), and it removes the fallback's misattribution to the DFS root. No behavior change on any reachable path.

Still not `expect()`. Exit codes here are a documented stable contract (`0` ok, `1` validation failure, `2` stale, `3` I/O); a panic exits 101 and aborts the gate outside it. Your own third option, "emit a proper `Violation` instead of masking the bad state", is what this now does, and it is strictly more useful to a consumer than a backtrace.

## Not acted on

**"`V-014` embedded in assert strings ... must be emitted through `out`."** It is, and always was. Fifteen lines below the assert, in the same arm:

```rust
out.push(error("V-014", message, at_path));
```

The assert string only labels which check the assertion belongs to.

**"Coupling-gate violation: no `spec.md` modified."** Handled, and the gate agrees. The PR body carries a cited `Spec-Drift-Waiver:` for exactly this path, and `self-governance (compile · index · lint · couple)` passes on this PR. A `debug_assert!` changes no claim in any of the nine specs owning `compile.rs`, so amending one would be the mechanical refresh that `.claude/rules/adversarial-prompt-refusal.md` forbids. The waiver is the sanctioned instrument.

**"Confirm `V-014` exists under `specs/`."** It does: `specs/033-dependency-cycle-refusal/spec.md`, merged in #72 and `approved`. `registry show 033-dependency-cycle-refusal` resolves, and `lint --fail-on-warn --fail-on-info` is 0/0/0.

The last two, plus the `016-short-id-resolution` flag on #72, are the same failure mode: asserting a corpus fact is missing when only the diff was in view. Worth checking the repo state before raising that class of finding.

bartekus:
Round three: two taken, two declined. Pushed as `fd7dde8`. Both changes are zero behavior change on any reachable path.

## Taken

**`unwrap_or(BLACK)` fails in the wrong direction.** Agreed, and this is the better half of the point. BLACK would skip the start node and miss a cycle rooted there, and a silent false negative is the one outcome a gate must never produce. It now defaults to `WHITE`, so an unreachable-but-reached branch walks the node instead, which is harmless when redundant. I did not add an "internal error" violation: that would mean minting a code no spec defines, which is a governance problem rather than a fix.

On "the code that seeds `ids` isn't shown": it is, four lines above the loop. `ids` is collected from these same `records`, `color` is built from `ids`, and no later insert can add a key because every `child` is filtered through `ids.contains(d)`. The branch is provably unreachable by local inspection of one function.

**`V-014` in the assert text.** Fair. A panicking debug test binary would put a real violation code on stderr where a tool counting gate errors could pick it up. The messages now read `dependency-cycle detector: ...`, naming the check without emitting the code.

## Declined

**"Early return drops subsequent cycles ... latent bug if the spec requires complete enumeration."** The spec requires the opposite. `specs/033-dependency-cycle-refusal/spec.md` line 105: "**One cycle per compile.** The first cycle found is reported and the walk stops." The rationale is in the same section and in the function's doc comment: the path names every spec on the cycle, and breaking it is what reveals any other. This is specified behavior, not an oversight.

**"Message shape change is an observable gate-output change, so the spec should have been updated."** The premise does not hold. The shape changes only on the branch that is unreachable while the invariant holds; on every reachable path `message` and `at_path` are byte-identical to `main`. No corpus can produce different gate output from this PR, so there is nothing for a spec to record. The cited `Spec-Drift-Waiver:` covers the path, and `self-governance (compile · index · lint · couple)` passes here.

## A note on the review loop

Across three rounds on two PRs, six of eleven findings have been the same failure mode: concluding a fact is missing because it is outside the diff. `V-014` not emitted through `out` (it is, fifteen lines below the assert), `016-short-id-resolution` may not exist (it does, and is approved), `V-014` may have no spec (it has, `033`, merged in #72), the seeding of `ids` "isn't shown" (it is, four lines up), and twice a coupling-gate violation that the passing `self-governance` job and the waiver in this body already answer.

The two genuinely useful findings, release-build silence and the BLACK fallback direction, both came from reasoning about the code that *was* in view. That is where this reviewer is strong. Checking repo state before asserting an absence would remove most of the noise.

bartekus:
Round four: no code change. One claim asserts a bug that does not exist, and I want the record straight on it rather than leaving it to look unanswered.

## "Walking an uninitialized node leaves broken bookkeeping"

> if `color` is missing a key, the DFS proceeds with no `color` entry for that node ... any child that already set its own color to `GREY` while this ghost node was on the stack may have that color cleared by the wrong pop

There is no ghost node. Two lines after the check, before the traversal loop begins:

```rust
if color.get(r.id.as_str()).copied().unwrap_or(WHITE) != WHITE {
    continue;
}
let mut stack: Vec<(&str, usize)> = vec![(r.id.as_str(), 0)];
color.insert(r.id.as_str(), GREY);   // <- the node is initialized on entry
```

The node enters the map as `GREY` on push, exactly like every other node. So `unwrap_or(WHITE)` is a walk with ordinary bookkeeping, not broken bookkeeping.

The second half does not hold either. There are three `color.insert` sites, at lines 428, 431 and 440, and each writes only its own key: `r.id` on entry, `node` on pop (`node` being the node popped), `child` on push. No pop can clear a child's color. There is no mechanism for the described corruption.

## Already answered

**Only one cycle reported.** Specified, not a defect: `specs/033-dependency-cycle-refusal/spec.md` line 105, "**One cycle per compile.** The first cycle found is reported and the walk stops." Turning `return` into `continue` would contradict an approved spec. The fallback arm does not "silently exit on a spurious hit": it is reached only when `child` is `GREY`, which proves a real cycle, and it pushes a `V-014` before returning.

**"Silent data loss" in the None branch.** It emits a full `V-014` violation through `out`. That is the opposite of silent. `log::warn!` is not an option regardless: `spec-spine-core` has no `log` or `tracing` dependency, by design, because core is IO-light.

**No `spec.md` touched.** This round contradicts itself on this point. The compliance section claims "the observable behavior of violation V-014 (cycle path computation, error message format, which spec path is reported)" changed; the Minor section immediately below concludes `at_path` is "Equivalent in the happy path; only diverges in the old `unwrap_or(0)` fallback that no longer exists". The second reading is correct, and it applies to the message too: on every reachable path this branch is byte-identical to `main`. No corpus produces different gate output, so no spec claim changes. The waiver in the PR body covers the path and `self-governance` passes.

## Closing the loop here

Fifteen findings across three rounds on this PR. Three were real and are fixed: release-build silence in the GREY arm, the `BLACK` fallback failing toward a false negative, and the `V-014` string in assert text. The rest were absences inferred from a partial diff, or points already answered in the thread. This round adds a concrete algorithmic claim that the source contradicts.

I am not making further changes on automated feedback here. A human reviewer overrules this at any time.

PR #74, opened 2026-09-03
chore(025,026): mark two shipped specs `implementation: complete`

bartekus:
Ledger drift: two approved specs still declare `implementation: pending` while
their code has been on `main` since v0.6.0.

| spec | shipped in | evidence in the tree |
|---|---|---|
| `025-unresolved-unit-severity` | `bea2fc3` (#32) | `W-001` / `W-002` emitted from `crates/spec-spine-core/src/index.rs` |
| `026-resolution-discovery-fixes` | `302fef9` (#33) | the `## tag:` handling in `sections.rs` cites "spec 026 D2" by name |

`registry list` was therefore reporting two shipped specs as unbuilt.

## Not cosmetic

025 is the spec that defines the severity tiers keyed on this very field. Its
own normative text (the severity table, and "an unresolved unit whose owning
spec is `status: draft` OR `implementation: pending` MUST be emitted as a
warning") makes the marker downgrade unresolved owning units from `I-0xx`
errors to `W-001`.

Flipping it to `complete` therefore **tightens** the classification for these
two specs. That was verified rather than assumed: after the change `index`
reports 0 error diagnostics and `lint --fail-on-warn --fail-on-info` is 0/0/0.

## Scope

Only the frontmatter line changed in each file. 025 discusses `implementation:
pending` throughout its normative body, so a naive replace would have corrupted
the spec that defines the rule; the edit is anchored to the frontmatter block
and the nine body occurrences are untouched.

## Verification

| check | result |
|---|---|
| `compile --check` | fresh, 34 shards |
| `index check` | fresh |
| `index` | 4 packages, 34 mappings, 0 error diagnostics |
| `couple --base main --head HEAD` | OK, 2 paths, no drift |
| `registry list` implementation tally | 33 `complete`, 1 unset (`000`, which has no such field) |

Each spec's own `spec.md` is the only authored file that changes, so the
coupling gate self-clears. No waiver needed.

### DIFF
--- a/.derived/codebase-index/by-spec/025-unresolved-unit-severity.json
+++ b/.derived/codebase-index/by-spec/025-unresolved-unit-severity.json
@@ -101,5 +101,5 @@
     "specStatus": "approved"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "30d0773f486bf48ffec22b0fd38a5283905a75039644bae6551a1c9f58b1cd7c"
+  "shardHash": "e028023928cdd8dfed4cb53ae7a919037af033d2c2ad44a3595777d620b94126"
 }
--- a/.derived/codebase-index/by-spec/026-resolution-discovery-fixes.json
+++ b/.derived/codebase-index/by-spec/026-resolution-discovery-fixes.json
@@ -84,5 +84,5 @@
     "specStatus": "approved"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "1b8eeafa462b73fc8f223ec3427767d6fbdc9c36ee5008f5f7cfd6e3ef2be483"
+  "shardHash": "80e0ddbd50782c65ff1ae253698ab5c1327f6132459118ddbb0b54595bc1a187"
 }
--- a/.derived/spec-registry/by-spec/025-unresolved-unit-severity.json
+++ b/.derived/spec-registry/by-spec/025-unresolved-unit-severity.json
@@ -44,7 +44,7 @@
       }
     ],
     "id": "025-unresolved-unit-severity",
-    "implementation": "pending",
+    "implementation": "complete",
     "kind": "tooling",
     "owner": "The spec-spine Authors",
     "references": [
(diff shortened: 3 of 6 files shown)


bartekus:
The blocking finding rests on a misreading of what the coupling gate requires, and taking it would make the ledger uncorrectable.

## The gate governs changed code, not past code

> coupling-gate requires spec and code to move together in the same change

It requires that a **changed** spec-claimed code path move together with its owning spec. No code path changes in this PR. `git diff --stat` is two `spec.md` frontmatter lines and four derived shards, and `self-governance (compile · index · lint · couple)` passes on this PR precisely because there is nothing for `C-001` to check.

The review's scenario 1 is what happened, and it is not a violation:

> Implementation landed in a prior PR and this is a trailing status update

Both prior PRs carried their spec and their code together, as the gate demanded at the time. What they got wrong was one frontmatter field. Requiring the implementation to be re-landed to correct that field would mean a stale marker can never be fixed: the code cannot move again, so under this reading the ledger would be permanently wrong. That inverts the purpose of the gate, which exists to keep the ledger true.

## The evidence was in the PR body

> The shardHash changes suggest tracked implementation files changed, but those files are absent from the diff.

The shard hashes changed because the shard embeds `implementation`, which is what this PR edits. No tracked source file changed; `index check` is fresh both before and after.

The implementations are on `main` and were named in the description with their commits:

| spec | landed | verifiable now |
|---|---|---|
| 025 | `bea2fc3` (#32) | `grep -o '"W-00[12]"' crates/spec-spine-core/src/index.rs` |
| 026 | `302fef9` (#33) | `sections.rs:139` cites "spec 026 D2" by name, beside the `## tag:` handling |

## Not cosmetic, and verified in the other direction

Worth restating, since the review treated this as metadata with no behavior: 025 is the spec that **defines** the severity tiers keyed on this field. Its own normative text makes `implementation: pending` downgrade an unresolved owning unit from an `I-0xx` error to a `W-001` warning. Flipping to `complete` therefore **tightens** the gate for these two specs, which is why the description records that `index` reports 0 error diagnostics and `lint --fail-on-warn --fail-on-info` is 0/0/0 afterwards.

The correct reading of `implementation: complete` here is the literal one: the code that delivers the claimed behavior is present on `main` and can be pointed at. It is.

PR #75, opened 2026-09-03
ci(ai-review): give the reviewer base-commit repo context, and bound its claims

bartekus:
The AI reviewer sees only a diff, so it cannot observe what the repository
already contains. It kept reporting things as missing that are present a few
lines outside the hunk.

Across #72 and #73, **six of fifteen findings were this one failure mode**:

| claimed | actual |
|---|---|
| "`V-014` is never emitted through `out`" | it is, fifteen lines below the hunk |
| "`016-short-id-resolution` may not exist" | it exists and is `approved` |
| "`V-014` has no spec under `specs/`" | `033`, merged in #72 |
| "the code seeding `ids` isn't shown" | four lines above the loop |
| unwaived coupling-gate violation (x2) | a cited waiver, and a passing `self-governance` job |

## What this does not do

It does **not** let claude read the PR tree. Running from an empty cwd with
input on stdin is a deliberate security control, documented in the step: it
keeps a contributor-controlled checkout, which a malicious PR could seed with a
`.claude/settings.json` granting Bash or `bypassPermissions`, off claude's
settings-discovery path.

So the context is assembled **by the runner** and piped in on the same stdin
channel. Every fact is read from `BASE_SHA`, the already-merged base commit,
never from the PR head. claude still receives bytes and no filesystem access,
and the agent-injection seam stays closed.

## What reaches stdin now

1. **REPO CONTEXT** (trusted, from `BASE_SHA`): every spec directory, and every
   tracked path except `.derived/` shard noise.
2. **PR METADATA** (contributor-controlled): only lines matching
   `^\s*Spec-Drift-Waiver:`, never the whole body. A narrow structured
   extraction imports the one fact the reviewer was missing without importing
   arbitrary contributor prose into the prompt.
3. **PR DIFF** (contributor-controlled), as before.

The prompt marks 2 and 3 as data and forbids following instructions inside them.

## Bounding the claims

The prompt now states what the reviewer may and may not conclude:

- Absence is not observable from a diff. Do not report a file, spec, symbol or
  declaration missing unless REPO CONTEXT shows it absent.
- Unchanged code beside a hunk is not shown, so "never emitted" or "never
  initialized" needs better evidence than not seeing it.
- Check the waiver list before reporting a coupling-gate violation.
- An unreachable branch is a design note, not a bug. Say which.
- Prefer few well-evidenced findings over many speculative ones; cite file and
  line.

## One incidental fix

The manifest uses `awk 'NR<=800'` rather than `head -800`. `head` closes the
pipe once satisfied, and under the step's `set -o pipefail` that SIGPIPE would
fail the step as soon as the tree outgrew the cap. The tree is 239 files today,
so this is latent rather than live, but it is the kind of thing that fails
confusingly a year from now.

## Verification

Run locally against a real base sha, since a workflow only proves itself in CI:

| check | result |
|---|---|
| YAML parses | 12 steps (was 11) |
| context step, real `BASE_SHA` | 280 lines, all 34 spec directories listed |
| waiver extraction, multi-line body | the `Spec-Drift-Waiver:` line only |
| no-waiver body, under `set -euo pipefail` | `(none declared)`, step survives |
| empty / absent body | `(none declared)`, step survives |
| `compile --check` / `index check` / `lint` | fresh / fresh / 0-0-0 |
| `couple --base main --head HEAD` | OK, 0 paths (`.github/` is on the bypass floor) |

`DIFF_SIZE_CAP` still measures the diff alone, so the added context does not
push PRs over the skip threshold.

This PR is its own first test: the review that runs on it will be the first one
with the context block.

### DIFF
--- a/.github/workflows/ai-pr-review.yml
+++ b/.github/workflows/ai-pr-review.yml
@@ -120,6 +120,70 @@ jobs:
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: git diff "${BASE_SHA}...${HEAD_SHA}" > /tmp/pr-diff.txt
 
+      # A diff cannot show what the repository already contains, so a diff-only
+      # reviewer keeps "discovering" that things are missing when they are
+      # present a few lines outside the hunk. Observed repeatedly on #72/#73:
+      # "V-014 is never emitted through `out`" (it is, fifteen lines below the
+      # hunk), "016-short-id-resolution may not exist" (it does), "V-014 has no
+      # spec" (033), "the code seeding `ids` isn't shown" (four lines up), and
+      # twice a coupling-gate violation already answered by a waiver.
+      #
+      # The fix is NOT to let claude read the PR tree. Running from an empty cwd
+      # with the diff on stdin is a deliberate security control (see "Run AI
+      # Review" below): it keeps a contributor-controlled checkout off claude's
+      # settings-discovery path. So the context is assembled HERE and passed on
+      # the same stdin channel, and every fact is read from BASE_SHA, which is
+      # the already-merged base commit, never from the PR head.
+      - name: Build repo context (from the base commit only)
+        if: steps.diff.outputs.skip != 'true' && steps.ctx.outputs.fork != 'true' && steps.ctx.outputs.dependabot != 'true'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          set -euo pipefail
+          {
+            echo "===== REPO CONTEXT (read from base commit ${BASE_SHA}; trusted) ====="
+            echo
+            echo "## Specs that exist at base"
+            git ls-tree -d --name-only "${BASE_SHA}:specs" 2>/dev/null || echo "(no specs/ directory)"
+            echo
+            echo "## Tracked files at base (paths only)"
+            # .derived/ is machine-generated shard noise; the shard names add
+            # nothing a reviewer needs and would crowd out the source tree.
+            # `awk NR<=N` rather than `head -N`: head closes the pipe when it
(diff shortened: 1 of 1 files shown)


bartekus:
Declining this one, and the reasoning matters more than the verdict.

## The branch is unreachable

`grep -v '^\.derived/'` exits 1 only when *every* tracked path at base is under `.derived/`. That cannot happen in a repository running this workflow: `.github/workflows/ai-pr-review.yml` is itself tracked and is not under `.derived/`, so at least one line always survives. The file that would abort the step is the file doing the aborting.

## The suggested fix would trade a loud failure for a silent wrong one

```bash
git ls-tree ... | { grep -v '^\.derived/' || true; } | awk ...
```

`|| true` swallows grep's exit **2** as well, which is a genuine error (an unreadable tree, a bad pattern), not an empty result. And in the empty case it would emit an empty manifest into a prompt that tells the reviewer the list is complete and that it may therefore conclude a path is absent.

That is exactly the failure this PR exists to remove, and it is the same defect you correctly identified in the 800-line cap one comment earlier. There the fix was to make the shortfall announce itself, which is why that finding was taken and this one is not. An aborted step is visible in the checks UI; a silently empty context block is not.

If the branch ever did become reachable, the correct fix would be a sentinel like the truncation one, not suppression.

## Also worth noting

The comparison drawn to the `head`→`awk` substitution is not symmetric. That change addressed SIGPIPE, where the failure is *timing-dependent* and fires whenever the tree outgrows the cap: reachable, and silent about why. This is a fixed-point condition that the repository's own contents rule out.

The other two findings in this round were taken and are pushed: the truncation sentinel now names the real count and revokes the absence licence, and `PR DIFF` has a closing marker. Your prompt-injection note was also taken, and I appreciated that it was graded as residual risk rather than a vulnerability: `=====` runs in extracted waiver text are now neutralized to `[=]`, which closes the structural half while the prompt's "data, not instructions" rule keeps covering the words.

PR #76, opened 2026-09-03
feat(034): `references` seeds no implementing path (C-001 ownership leak)

bartekus:
`standards/spec/contract.md` states the rule in one line: eight typed edges,
"`references` is the only non-owning one (**the coupling gate ignores it**)."
The indexer did not implement that.

Building each spec's traceability mapping, it seeded `implementingPaths` from
every resolved unit location, holding the unit's `ownership` flag and not using
it. A `references` edge produced a claim indistinguishable from an `establishes`
one.

## Why that matters in four places

`implementingPaths` is read as an authority claim by:

| reader | reads a listed path as |
|---|---|
| `couple.rs::owners_for_path` | whole-file `C-001` ownership |
| `index.rs::owners_of_unit` | ownership of a file unit |
| `orphaned_specs` | "this spec claims something" |
| coverage (spec 032) | a claimed source file |

The `resolvedUnits` loops sitting beside the first two check `ownership`
correctly. The `implementingPaths` loops did not, so one bad seed bypassed a
guard that already existed.

## The defect is live on this corpus

`024-index-sharding` references `specs/020-derived-artifact-merge-driver/spec.md`
for context. So editing spec 020's **own** `spec.md` was refused:

```
$ spec-spine couple --paths-from <(echo specs/020-derived-artifact-merge-driver/spec.md)
C-001 'specs/020-derived-artifact-merge-driver/spec.md' changed without an
authoring edit to any owning spec (024-index-sharding)
exit 1
```

A spec could not be edited without touching an unrelated spec that has no
authority over it, or filing a waiver. Citing a document conferred power over
it. After this change the same command is `OK: 1 path(s) checked, no drift`.

## The fix

One `if *ownership` around the seed, in the indexer, at the single place the
field is built. Not in the four readers: that would repeat one predicate four
times and leave the emitted index still asserting a claim that is not true.
`implementingPaths` means "paths this spec implements", and a cited file is not
one.

**Provenance is preserved.** The unit stays in `resolvedUnits` with
`ownership: false` and its resolved locations, so "which files does this spec
cite" is still answerable from the index. Nothing recorded stops being recorded;
a claim that was never true stops being asserted.

## Blast radius

No DTO field added, removed or retyped, and no JSON Schema change, so
`INDEX_SCHEMA_VERSION` does not move. Eight index shards restamp: exactly the
specs carrying `references` edges (023, 024, 025, 026, 027, 028, 031, 032).

Orphans and coverage are unchanged here, because every referenced target in this
corpus is a document or a `spec.md`, not a source file. A spec whose only units
were `references` would now report as orphaned, which is correct: it implements
nothing.

## Tests

Three, in `tests/index.rs`. Verified non-vacuous by reverting the fix: the first
and third fail without it, and the second is a provenance regression guard that
should pass either way.

1. An owning edge and a `references` edge to two equally-real files: only the
   owning path is claimed.
2. The citation survives in `resolvedUnits` with `ownership: false`.
3. End to end: a spec that only references another spec's `spec.md` is not its
   owner, so editing that `spec.md` alone clears the gate.

## Verification

| check | result |
|---|---|
| `compile --check` | fresh, 35 shards |
| `index check` | fresh |
| `couple --base main --head HEAD` | OK, 3 paths, no drift |
| `index orphans` | unchanged, exit 0 |
| `index coverage` | unchanged, 59/64 |
| `cargo fmt` / `clippy -D warnings` / `test --workspace` | clean / clean / all pass |

Spec 034 is filed here as `draft`.

## Note on merge order

This PR and #74 both restamp the `025` and `026` index shards, so whichever
lands second needs a regenerate. That is the same-shard case spec 020's merge
driver exists for; a plain `compile && index` on the merged tree resolves it.

## Out of scope, deliberately

- `couple.rs` hardcodes `"specs/"` at lines 337 and 422 while `Config::specs_dir`
  is configurable, which breaks the gate for an adopter who moves their corpus.
  Real, but it is about locating a spec's authoring file, not about which edges
  confer ownership.
- Whether a `spec.md` should be ownable by another spec at all. `amends` and
  `supersedes` deliberately transfer authority, and this PR does not revisit
  that.

### DIFF
--- a/.derived/codebase-index/by-spec/023-ledger-seal.json
+++ b/.derived/codebase-index/by-spec/023-ledger-seal.json
@@ -46,10 +46,6 @@
       {
         "path": "crates/spec-spine-types/src/lib.rs",
         "source": "spec-edge"
-      },
-      {
-        "path": "docs/design/00-architecture.md",
-        "source": "spec-edge"
       }
     ],
     "resolvedUnits": [
--- a/.derived/codebase-index/by-spec/024-index-sharding.json
+++ b/.derived/codebase-index/by-spec/024-index-sharding.json
@@ -108,18 +108,6 @@
       {
         "path": "crates/spec-spine-types/tests/dtos.rs",
         "source": "spec-edge"
-      },
-      {
-        "path": "docs/design/00-architecture.md",
-        "source": "spec-edge"
-      },
-      {
-        "path": "specs/012-index-hash-slices/spec.md",
-        "source": "spec-edge"
-      },
-      {
-        "path": "specs/020-derived-artifact-merge-driver/spec.md",
-        "source": "spec-edge"
       }
     ],
     "resolvedUnits": [
--- a/.derived/codebase-index/by-spec/025-unresolved-unit-severity.json
+++ b/.derived/codebase-index/by-spec/025-unresolved-unit-severity.json
@@ -24,10 +24,6 @@
       {
         "path": "crates/spec-spine-types/tests/dtos.rs",
         "source": "spec-edge"
-      },
-      {
-        "path": "docs/design/00-architecture.md",
-        "source": "spec-edge"
       }
     ],
     "resolvedUnits": [
--- a/.derived/codebase-index/by-spec/026-resolution-discovery-fixes.json
+++ b/.derived/codebase-index/by-spec/026-resolution-discovery-fixes.json
@@ -20,10 +20,6 @@
       {
         "path": "crates/spec-spine-core/tests/index.rs",
         "source": "spec-edge"
-      },
-      {
-        "path": "docs/design/00-architecture.md",
-        "source": "spec-edge"
       }
     ],
     "resolvedUnits": [
--- a/.derived/codebase-index/by-spec/028-references-provenance-derived-at.json
+++ b/.derived/codebase-index/by-spec/028-references-provenance-derived-at.json
@@ -19,10 +19,6 @@
       {
         "path": "crates/spec-spine-types/tests/grammar.rs",
         "source": "spec-edge"
-      },
-      {
-        "path": "docs/design/00-architecture.md",
-        "source": "spec-edge"
       }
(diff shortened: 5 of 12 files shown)


bartekus:
Both points checked against the tree rather than assumed. Neither needs a change, and the first was worth verifying.

## "Verify `orphaned_specs` was never wired to an error code"

Correct instinct, and it holds. `orphaned_specs` is a plain `Vec<String>` on the traceability report. Its only consumers are the `index orphans` projection and `orphans_json` (`core/src/lib.rs:231`); no code path turns it into a diagnostic, and `index orphans` exits 0 on this corpus with orphans absent or present. So "reported, never blocking" is established behavior, not an assumption the test smuggles in.

On the framing though: the test asserts `diagnostics.errors.is_empty()`. If orphan detection *did* emit an error, that assertion would **fail**, which surfaces the problem rather than masking it. The test is correct in either world; it just also documents which world we are in. That is why it is written as an assertion instead of a comment.

## "`docs/design/00-architecture.md` has no owner after this PR"

Accurate observation, and it is the intended end state rather than a gap. No spec ever *owned* that file. Eight specs cited it on a `references` edge, and the bug was that citing conferred ownership. Removing a claim that was never legitimate leaves the file where it always belonged.

Nothing downstream changes, verified on the branch:

| check | result |
|---|---|
| `couple --paths-from docs/design/00-architecture.md` | `OK: 0 path(s) checked` (the `docs/` bypass floor, unchanged) |
| `index coverage` | `59/64 specifically claimed, 5 floor-only, **0 unclaimed**` (identical to `main`) |

Coverage never counted it: it classifies *source* files inside discovered packages, and a design document is not one. So there is no "unclaimed files" surface for it to appear on, and editing it does not trip `C-001`, before or after.

If someone later wants a spec to genuinely own that document, the instrument is an owning edge, which is exactly the distinction this PR restores.

PR #77, opened 2026-09-03
feat(035): a closed reader ends the CLI cleanly instead of panicking

bartekus:
Exit codes are a stable contract (`0` ok, `1` validation failure, `2` stale, `3`
I/O), mapped in exactly one place via `Error::exit_code()`. The CLI could still
exit **101** through a path that never reached that mapping:

```
$ spec-spine registry list --json | head -c 50
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
$ echo ${PIPESTATUS[0]}
101
```

`println!` unwraps its write, so a reader that stops early panicked the process,
and under `set -o pipefail` that failed the surrounding script. Reading part of
a stream is not an error: `head`, `less`, `grep -q` and any pager all do it.

## The fix

All 39 CLI stdout writes go through one helper that classifies the write instead
of unwrapping it:

| outcome | when | process |
|---|---|---|
| written | success | continue |
| reader gone | `ErrorKind::BrokenPipe` | exit `0` |
| failed | any other I/O error | report on stderr, exit `3` |

`0` for a closed reader is the shell convention (`producer \| head` succeeds).
`3` for a real failure is the code the contract already assigns to I/O, so
`main.rs` stays the single authority on what each code means.

Output is byte-for-byte identical: the helper writes the same formatted line to
the same handle. No artifact, hash or golden test moves. Only the behavior when
a write *fails* changes.

**Stderr is deliberately unchanged.** `eprintln!` stays at every diagnostic
site: that output is small, and a process whose stderr has gone has no channel
left to report the fact.

## Why not the two shorter fixes

**Restoring the default `SIGPIPE` disposition** is the usual remedy and is
unavailable: it needs an `unsafe` libc call, and `unsafe` is `forbid`-en
workspace-wide in `[workspace.lints]`. Taking a third-party crate to perform the
same `unsafe` behind a safe wrapper would honor the lint's letter against its
intent, and would add a dependency to the determinism surface for one signal.

**A panic hook or `catch_unwind`** would have to recognize the broken-pipe panic
by its message text. That string belongs to the standard library and can change
on a toolchain bump, so the guard would fail silently; it also leaves `println!`
panicking underneath. `ErrorKind::BrokenPipe` is stable API.

## Tests

- The classifier directly: `Ok` is written, `BrokenPipe` is reader-gone, other
  kinds are failures.
- End to end: a fixture emitting more than a pipe buffer (30 specs with an 8 KiB
  summary, so it exceeds the 64 KiB buffer on any platform) whose reader closes
  after 32 bytes. Asserts exit `0`, not `101`, and no panic on stderr.

Verified non-vacuous by reverting the fix in `cmd_registry.rs`, which reproduces
`left: Some(101)`.

## Verification

| check | result |
|---|---|
| `compile --check` | fresh, 35 shards |
| `index check` | fresh |
| `couple --base main --head HEAD` | OK, 12 paths, no drift (self-clears, no waiver) |
| `cargo fmt` / `clippy -D warnings` / `test --workspace` | clean / clean / 23 suites pass |
| `registry list --json \| head -c 50` | exit `0` (was `101`) |

Spec 035 is filed here as `draft`. Its `extends` edges name `001` for all ten
modified files because this is crate-wide output plumbing: no command's meaning
changes, only the call it makes to write a line. `out.rs` is `establishes`.

## A note on provenance

This is the same argument spec 033's review turned on, where an `expect()` was
refused because 101 sits outside the documented contract. The CLI was doing on
its most ordinary path what the library had just declined to do on an
unreachable one.

## Merge order

Independent of #74, #75 and #76: no shared shards, and the only overlap with
#76 is that both add a new spec directory.

### DIFF
--- a/.derived/codebase-index/by-spec/035-stdout-closed-reader.json
+++ b/.derived/codebase-index/by-spec/035-stdout-closed-reader.json
@@ -199,5 +199,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "1e37bd9e924c0d4203cda1028f5af0cd267403717202611bdd24fca061dd7fc6"
+  "shardHash": "5a321266b8796d029344f93730dcfab6ccf37c5ff924eaf4e8e82bb82557344a"
 }
--- a/.derived/codebase-index/by-spec/035-stdout-closed-reader.json
+++ b/.derived/codebase-index/by-spec/035-stdout-closed-reader.json
@@ -1,0 +1,203 @@
+{
+  "mapping": {
+    "dependsOn": [
+      "001-compile-registry",
+      "002-registry-query"
+    ],
+    "implementingPaths": [
+      {
+        "path": "crates/spec-spine-cli/src/cmd_attest.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_compile.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_couple.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_index.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_init.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_lint.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/cmd_registry.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/main.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/out.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-cli/src/verify_attestation.rs",
+        "source": "spec-edge"
+      },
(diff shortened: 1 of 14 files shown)


bartekus:
Real bug, and a consequential one. Fixed in the push above.

`line.find(mac)` returns only the first occurrence, and `println!(` occurs inside `eprintln!(`. So on

```rust
eprintln!("{}", x); println!("{}", y);
```

the search matched at offset 1 inside the stderr call, saw `e` before it, marked the line stderr, and never examined the real stdout call behind it.

That matters more than an ordinary false negative because of what this test is for. Spec §3.5(3) leans on it as the proof for the block path, which cannot be exercised by a pipe-breaking test: on any corpus small enough to build in one, `index render` output fits inside a pipe buffer, so such a test could never fail. A guard with a hole is worse than no guard, because the guarantee is then asserted on evidence that does not hold.

The scan now walks every occurrence. I also extracted it to `panicking_stdout_macros` and unit-tested the scanner directly rather than only through the directory walk, so its own correctness is pinned rather than assumed: stderr-only lines clear, stdout-only lines flag, comments skip, and a stderr call no longer masks a stdout one behind it. Reverting the loop to the old find-first form fails that test on exactly the mixed line, which is how I checked it is not vacuous.

Two rounds ago this same test was `read_dir`-shallow; now it is recursive and occurrence-complete. Both were your findings and both were right.

One limitation stands deliberately: the scanner does not parse Rust, so `print!` inside a string literal would be a false positive. That direction fails safe (a spurious failure a human clears in seconds, never a missed panic path), and a parser here would cost more than the property is worth.

PR #89, opened 2026-09-06
docs(037-039): file wave 1 of the agentic-builder substrate

bartekus:
## Summary

Files three draft specs plus the design note they cite, after analysing what an
external provider-agnostic autonomous builder needs from spec-spine when
spec-spine is its adjudicator rather than one of its agents.

The headline result of that analysis is that **most of the substrate asks are
already shipped**: `--repo`, `layout.specs_dir` (036), `frontmatter.extra_known_keys`
(013), the `0/1/2/3` taxonomy, `couple --paths-from`, `attest`/`verify-attestation`
(023), `depends_on` + cycle refusal (033), the complete JSON facade, and spec 025's
W-band downgrade that makes a draft spec's not-yet-existing units legal. That is
why wave 1 is three small, additive specs and not a redesign.

| Spec | What it proposes |
|---|---|
| `037-machine-readable-verdicts` | `--json` on the six adjudicating verbs (`couple`, `lint`, `compile --check`, `index check`, `attest`, `verify-attestation`), behind one versioned envelope whose `report` member is byte-identical to the corresponding facade output. Exit codes unchanged; failures reported inside the envelope. |
| `038-registry-plan-ready-set` | `registry plan`: a ready/blocked partition in deterministic topological order, `blockedBy` naming the reason. |
| `039-declared-state-dir` | `layout.state_dir`: one declared root the gates recognise (bypassed, uncounted, unscanned, unhashed) and spec-spine refuses to govern. Unset by default. |

`docs/design/02-agentic-builder-substrate.md` is the durable record: the boundary
rule (would an adopter who never runs the builder still want it?), the nine asks
already satisfied, eight gaps, four constraints on wave 2, and the wave plan.

No engine code changes. All three specs are `status: draft` /
`implementation: pending`, so they declare territory without asserting anything is
built.

## Two things worth a reviewer's attention

**1. `038` amends `033`, deliberately.** Spec 033 §4 places "readiness, pinning,
and invalidation" outside spec-spine, reasoning that they are functions of merge
history and a run journal and so cannot be computed from `(config, file contents)`.
That reasoning is correct and 038 computes none of those three things: it computes
a strictly weaker, demonstrably pure partition over fields the corpus already
carries. The word "readiness" covers both, which is where the collision comes from.
038 therefore `amends` 033 to narrow the exclusion rather than quietly contradicting
it, and argues the distinction in its §4. **Spec 033's own text is unchanged.**
If the reviewer thinks the narrowing is wrong, 038 is the spec to push back on.

**2. The sharpest gap is deferred, not solved.** `implementation: complete` is
self-declared and `lint.rs` reads it nowhere, so nothing verifies it. 038 §3.4
states in terms that `plan` treats the field as a claim about intent and never as
evidence about code. The completion gate that would make it trustworthy is wave 2.
Until it lands, no consumer should route acceptance decisions on that field.

## Testing

Gate chain, run locally against this branch with the release binary:

```
compile                            0   40 specs, 0 warnings
compile --check                    0   spec-registry fresh, 40 shards
index coverage --fail-on-untraced  0   65/65 (100.0%), 0 floor-only, 0 unclaimed
couple --base origin/main --head HEAD   0   3 paths checked, no drift
cargo test --workspace --locked    0   all suites pass
```

Sharding behaved as spec 024 intends: filing three specs wrote exactly six new
shards and restaled nothing else, including spec 033's, whose `shardHash` is over
its own `spec.md` source bytes and is unaffected by an inbound `amends` edge.

Spec 037 claims `crates/spec-spine-types/src/verdict.rs`, which does not exist yet.
Because the spec is draft/pending this resolves to a `W-001` warning rather than a
blocking `I-0xx`, and `W-001` is outside `BLOCKING_CODES`, so `index check` stays
fresh. That is spec 025 working as designed.

## Incidental finding, not fixed here

`cmd_index.rs:134-143` iterates and counts only `diagnostics.errors`. The W-band
warnings land in the committed shard but no verb prints them, so `spec-spine index`
reports "0 error diagnostic(s)" in the same run that recorded a warning. Surfaced in
the artifact, not on the CLI. Left alone to keep this PR to filing; it strengthens
the case for 037 rather than competing with it.


bartekus:
Worked through all six review items. Three were valid and are fixed in 889f263; three are false positives, with evidence below.

## Fixed

**`state_dir` bounds (039 §3.2, §3.6).** The ancestor case is real and I have promoted it from a test bullet to a normative MUST: the value must not be, and must not be an ancestor of, `specs_dir` or `derived_dir`, rejected at config load. The reasoning is now stated too, because it is the sharpest argument for the rule: a `state_dir` of `.` makes every gate keep exiting 0 while adjudicating nothing, which is a silent stop rather than a refusal.

One correction to the finding itself: `specs_dir` was already covered. "The corpus root" is this repo's term for it, established by spec 036 ("`Config::layout.specs_dir` names the corpus root"). The wording was ambiguous enough to mislead a careful reader, so the fix now names both keys literally instead.

**`state_dir` §3.5 wording.** Agreed, "MUST NOT be removable by any list" was ambiguous. Replaced with what it meant: neither `bypass_prefixes` nor `resolver_exclusions` can cancel the key's effects, and setting both to cover the same path is not equivalent to declaring it.

**038 enrichment scope.** Fair, §4 was silent. `plan` emits ids only; `registry show` is the verb for the rest. Added the reason as well as the rule: inlining `implementation` would place a self-declared field beside computed ones, where a reader would reasonably assume both had been adjudicated.

## Not fixed, with evidence

**"038 `amends` 033 without updating 033."** This is the repo's established and deliberate pattern, not an oversight. `.claude/rules/adversarial-prompt-refusal.md` forbids editing an owning spec to accommodate a change made elsewhere, and the amendment mechanism exists so a predecessor's history stays intact. Both precedents behave identically: spec 025 `amends` 004 and did not touch `specs/004-codebase-index/spec.md`; spec 036 `amends` 005 and commit 46474a3 touched only `specs/036-configured-corpus-root/spec.md`. The bidirectional view is `registry relationships`, which is the typed read that makes the narrowing visible from 033's side without editing 033. Adding a forward pointer by hand would duplicate the edge in prose and would need maintaining every time an amendment lands.

**"`docs/design/02-...md` is unowned and will fail the coupling gate."** `docs/` is in the built-in bypass floor at `crates/spec-spine-core/src/couple.rs:38`. Empirically: `couple --base origin/main --head HEAD` exits 0 locally, and the `self-governance` job on this PR passed, which runs the same gate. `ownership: false` with `sourceField: references` is the correct and intended record for a `references` edge since spec 034.

**"`diagnostics` present in the 037 shard but absent in 038/039."** Optional by schema. `crates/spec-spine-types/schemas/codebase-index-spec-shard.schema.json:7` requires only `["schemaVersion", "shardHash", "mapping"]`. The conformance test asserts emitted JSON validates against the embedded schema and passed in `build · test · clippy`. The 037 shard carries `diagnostics` because it holds the `W-001` for the not-yet-written `verdict.rs`, which is spec 025 working as designed for a draft spec.

Gate after the fixes: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 2: two findings valid and fixed in d701ace, two already answered above and now settled empirically.

## Fixed

**`deferred` treated as `complete` (038 §3.1).** Good catch, and the spec was asserting the boundary without arguing it. `plan` does not ask "is this finished", it asks "should this be scheduled", and to both `complete` and `deferred` the answer is no. But `deferred` is the one value in the enum that is a *decision* rather than a report, so §3.1 now says that, and says why the scheduler must not reverse it: a `plan` that quietly returned a deferred spec to `ready` when a dependency landed would make deferral expire on its own, which is not what deferring something means. Un-deferring is a human edit, exactly as deferring was. Behavior unchanged; the reasoning is now written down, which is what the finding asked for.

**`attest --json` not covered by the out-of-scope rationale (037 §4).** Also fair: §4 argued that writing commands are excluded, then included one. Now distinguished. What `attest` writes *is* the verdict, so its stdout and its artifact agree by construction, and it does not regenerate the shards the next gate compares against, which is the specific coupling that makes `compile` and `index` awkward to consume mid-chain. On the stdout-as-error-indicator concern: `--json` is opt-in, so no existing caller of any of these commands changes behavior until it passes the flag. Noted in the same paragraph.

## Settled

**The `amends` coupling-gate concern.** The hypothesis was "if the tool evaluates `amends` the way `extends` is evaluated, this PR will fail the gate without a waiver". It does not, and the code is unambiguous: `couple.rs:317-330` uses `amends` in exactly one direction, to *widen* the owner set for a changed `<specs_dir>/<id>/spec.md` under the FR-005 strict-expansion guard, so that editing the amending spec can clear a violation on the amended spec's territory. There is no path by which an `amends` edge demands that the amended spec change. Confirmed empirically as well: `self-governance (compile · index · lint · couple)` passed on `889f263`, and `couple --base origin/main --head HEAD` exits 0 locally on every head this branch has had.

**`diagnostics` shape and the "spurious diff" prediction.** The prediction was that regenerating would rewrite whichever shards are currently wrong. Ran `index` twice more: zero modified files, `index check` fresh. The field is `skip_serializing_if`-empty, so its presence is content-driven, not shape-inconsistent: the 037 shard has it because it holds the `W-001` for the not-yet-written `verdict.rs`, and 038/039 have no diagnostics to carry. Requiredness was already settled at `codebase-index-spec-shard.schema.json:7`, which requires three keys and not this one. If the shape were unstable, the four determinism jobs across the release triples would be the ones to say so, and they pass.

Gate on `d701ace`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 3: four findings valid and fixed in 8725674. One rests on a rule this corpus does not have.

## Fixed

**`deferred` blocker named but absent from the output (038).** The best finding of the three rounds, and a real incoherence I introduced: §3.1 excludes `deferred` specs from both sets while §3.5 has them block their dependents, so a bare id could name a blocker the reader can find nowhere else in the document. Rather than make `deferred` non-blocking (it genuinely does block: a dependent is not workable because someone deferred its dependency), every `blockedBy` entry now carries the blocker's `state`:

```json
{ "id": "031-migration", "blockedBy": [ { "id": "030-round-trip", "state": "deferred" } ] }
```

`state` is the blocker's `implementation` value, or `unresolved` for a dangling `depends_on` target, which also folds §3.1's dangling-dependency rule into the same field. §3.1 now names the asymmetry explicitly instead of leaving a reader to discover it, and two tests cover it. This also repaired a bullet list that my previous commit had split in half.

**`kind` promised stable without enumeration (037 §3.3).** Correct, and unverifiable as written. The set is now enumerated as a table of the seven tokens against their `Error` variants and exit codes (`config`, `validation`, `not-found`, `stale`, `io`, `parse`, `schema`), with evolution bound to `VERDICT_SCHEMA_VERSION`: adding a kind is MINOR, renaming or removing one is MAJOR. Added the reason the token is not simply the Rust variant name lowercased, which is that an internal rename would then be a silent breaking change to an external contract with no gate to catch it.

**`state_dir` ancestor check was one-directional (039 §3.2).** Right, `specs_dir: specs` with `state_dir: specs/state` was accepted and would put a path inside a governed root and an ungoverned one simultaneously. Overlap is now refused in both directions. Refusing the configuration is cheaper than specifying which root wins, and leaves nothing to get wrong later.

**Closed-reader test named one verb under a universal requirement (037 §3.6).** Fixed: the test covers all six, and says so, so it cannot be discharged by a single `couple` case.

## Not fixed

**"A spec that extends a unit must depend on the spec that owns it."** That rule does not exist in this corpus, and adopting it would put twelve approved specs in violation. Measured across `specs/0[0-3]*`, the specs that `extend` a spec they do not list in `depends_on` are 010, 011, 012, 013, 014, 015, 018, 019, 023, 024, 028 and 032. Spec 032, for instance, extends `000-spec-spine-bootstrap` and `001-compile-registry` and lists neither.

That is by design rather than by neglect: `extends` is an authority relation (whose territory am I adding to) and `depends_on` is a build order, and collapsing them would make `depends_on` mechanically derivable and therefore redundant. The ordering concern is also moot in the concrete case, since `001-compile-registry` is `implementation: complete` and §3.1 excludes it from the partition entirely, so it can never be ordered against anything.

There was a real question underneath it, though, which is now answered in 038 §4: `plan` walks `depends_on` and nothing else, and does not infer order from `extends`, because doing so would silently redefine `depends_on` for all forty specs.

**`diagnostics` in the 038/039 shards.** Third time raised, and settled twice on evidence: optional at `codebase-index-spec-shard.schema.json:7`, which requires `["schemaVersion", "shardHash", "mapping"]`. The prediction that regeneration would produce a spurious diff was tested directly last round: two further `index` runs, zero modified files, `index check` fresh.

Gate on `8725674`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 4: one new finding and two useful test suggestions, all in a168b63. Two items were raised in earlier rounds and are settled.

## Fixed

**Cycle-error shape unspecified under `--json` (038 §3.2).** Correct, and the only genuinely new finding this round. §3.2 required `plan` to report a cycle as an error without saying what that error is, while §3.3 has `plan` riding inside 037's envelope, which does specify every error path. Now pinned: `Error::Validation` carrying the cycle path, exit `1`, `kind: validation`. It deliberately reuses spec 033's existing classification of the same defect rather than inventing a second vocabulary for it, and a test covers the branch by constructing a cycle directly, since `compile` refuses to produce one.

**`L-006` as an unverified forward claim (039 §3.6).** The claim is accurate today: `L-001`..`L-005` are the allocated codes in `lint.rs`, which I checked before writing it. But the point stands that a comment about a namespace does not defend itself between filing and implementation, so §3.6 now asserts that no two lint diagnostics share a code. That is a better invariant than the one the finding asked for, because it stays true regardless of how the band grows.

**Coverage denominator auditability (039 §3.6).** Adopted the suggested test: the denominator must shrink by exactly the number of files under the root, asserted as a count rather than a percentage, so declaring a state root moves the figure auditably rather than plausibly.

One correction to the premise: there is no `--min-coverage` threshold to be silently satisfied. `index coverage` takes `--json`, `--repo` and `--fail-on-untraced`, and the string does not appear anywhere in `crates/`. The gate is all-or-nothing at full coverage, not a tunable floor. The test is worth having regardless, which is why it is in.

## Settled in earlier rounds

**`diagnostics` in the 038/039 shards.** Fourth raising. Optional at `codebase-index-spec-shard.schema.json:7`, which requires `["schemaVersion", "shardHash", "mapping"]`. The field is `skip_serializing_if`-empty, so its presence is content-driven: the 037 shard carries it because it holds the `W-001` for the not-yet-written `verdict.rs`. The prediction that regeneration would produce a spurious diff was tested in round 2 and did not occur.

**A cross-reference in 033.** Third raising. Answered with the repo's precedents (spec 025 amends 004 without touching it; commit 46474a3 for spec 036 amends 005 and touches only its own `spec.md`) and with the mechanism at `couple.rs:317-330`, where `amends` only ever widens the owner set and never demands the amended spec change. `.claude/rules/adversarial-prompt-refusal.md` makes editing 033 to accommodate 038 the specific move to avoid, so it will not happen in this PR. The bidirectional view is `registry relationships`, which is the typed read that surfaces the narrowing from 033's side.

Gate on `a168b63`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 5: both findings valid and fixed in 42717ef. Thanks for withdrawing the `amends` and `W-001` items explicitly; that is useful signal.

## Fixed

**Internal contradiction in 038 §3.3.** Correct, and the sentence asserted both output formats at once. It is my defect, and the previous commit made it worse by adding a second reference to the envelope in §3.2.

Resolved to **bare**. `plan` emits its report the way `registry list`, `show`, `status_report` and `relationships` do, and does not ride inside 037's envelope. That is not a preference: 037 §4 already keeps the envelope off the read verbs on purpose, since wrapping a shipped read surface would break its consumers, and `plan` joining the other `registry` subcommands is what stops the group having two output shapes. The §3.2 cycle error now says the same thing: `Error::Validation`, exit `1` under the standing exit-code contract, reported like every other `registry` error.

Your structural point resolves with it: since the envelope does not apply, 037 is **not** a dependency of 038. §3.3 now says so, and that the two specs can land in either order.

**Overlap validation must read the resolved layout (039 §3.2).** The sharpest finding of the five rounds. `specs_dir` and `derived_dir` are both configurable, so the check has to compare against their resolved values, and my §3.6 matrix enumerated the default literals in a way an implementer could reasonably read as the rule. Under `specs_dir = "corpus"`, a `state_dir` of `corpus/state` must be refused on `corpus`, and a check written against the literal `specs` would clear it and quietly make every `spec.md` beneath it ungoverned.

This is precisely the defect spec 036 fixed in `couple.rs`, where two functions spelled the corpus root as a literal and the gate then searched a path the repo did not contain, so §3.2 now cites that precedent. The added test is deliberately the inverse of the default-layout case: under `specs_dir = "corpus"`, `corpus/state` is refused and `specs/state` is accepted, which fails against any implementation comparing with a hardcoded root.

## Settled

**`diagnostics` in the 038/039 shards.** Fifth raising, and the answer has not moved: optional at `codebase-index-spec-shard.schema.json:7`, whose `required` is `["schemaVersion", "shardHash", "mapping"]`. The field is `skip_serializing_if`-empty, so presence is content-driven; 037 carries it because it holds the `W-001` you correctly classified as expected. The "regenerate and the shards will change" prediction was tested in round 2: two further `index` runs, zero modified files.

Gate on `42717ef`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 6: the new finding is valid and fixed in a2a4eba. The other two are closed, and I am not going to re-answer them again.

## Fixed

**`L-006` is racy across parallel spec work (039 §3.4).** A good catch and a real hazard: a `pending` spec cannot hold a namespace, and the §3.6 invariant test only fires at implementation time, so a collision would surface in some other author's PR.

Rather than reserve the code in `lint.rs` (which would put implementation code in a docs-only PR), I made the reservation provisional and moved the binding to a rule instead of a number: the requirement is "the next free code in the `L-` band", and `L-006` is what that resolves to today. An implementer who finds the band has grown takes the next free code and updates the section rather than colliding, and the no-shared-codes invariant makes either outcome safe to discover. That removes the race at the spec level instead of mitigating it.

## Closed

**`diagnostics` in the 038/039 shards.** Sixth raising. The answer has been the same each time and is checkable in one line: `codebase-index-spec-shard.schema.json:7` declares `"required": ["schemaVersion", "shardHash", "mapping"]`. `diagnostics` is not in it. The field is `skip_serializing_if`-empty, so presence is content-driven, and 037 carries it only because it holds the `W-001` you have twice classified as expected. You name `crates/spec-spine-core/tests/conformance.rs` as the thing that would flag it: that test asserts emitted JSON validates against the embedded schema, and it has passed on all six heads of this branch.

**A cross-reference in 033.** Sixth raising, and round 5 explicitly withdrew it ("The `amends` edge is a spec-to-spec semantic relationship, not a file-ownership claim; the coupling-gate gates file units, not spec-spec edges. No coupling violation."). That analysis was right. The mechanism is at `couple.rs:317-330`, where `amends` only ever *widens* the owner set under the FR-005 strict-expansion guard and never demands the amended spec change. The precedents are spec 025 amending 004 and spec 036 amending 005, neither of which edited its predecessor. `.claude/rules/adversarial-prompt-refusal.md` makes editing 033 to accommodate 038 the specific move to avoid, so it will not happen in this PR, and no waiver is needed for a gate that passes.

One part of that finding is fair and worth recording rather than dismissing: the convention *is* under-specified in the corpus. It lives in the harness rules and in practice, not in spec 005 or the constitution, which is why it keeps looking like an open question from the diff alone. Writing it down belongs in a spec that amends 005, not in this PR, and I have noted it as a follow-up.

Gate on `a2a4eba`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Round 7: no changes. All three items resolve against the repo, and one of them would have made the PR worse if I had applied it. No new commit, so `a2a4eba` stands.

**`docs/design/02-...md` unowned (C-002 on future edits).** Your option (c) is the answer: this follows an already-established convention. `docs/` is in the built-in bypass floor at `crates/spec-spine-core/src/couple.rs:38`, so no path under it reaches the ownership check. The two existing design notes are in exactly the same position, which is checkable rather than assumed: nothing `establishes` `docs/design/00-architecture.md` or `docs/design/01-partial-supersession.md`. Making 02 an owned unit would be the deviation, and it would also make every future edit to the note require a spec edit, which is the opposite of what a design note is for. Empirically the gate agrees: `self-governance` has passed on all seven heads of this branch.

**`cli.rs` extended via `001` rather than `002`.** I checked this one rather than taking the suggestion, and the suggestion is wrong. Every spec in the corpus that extends `crates/spec-spine-cli/tests/cli.rs` attributes it to `001-compile-registry`: specs 010, 011, 012, 024, 031, 032 and 035, plus 037 and 038 in this PR. That includes `010-registry-query-projection-flags`, which is itself a `registry` feature and the closest analogue to 038. `002` would be the deviation here, not the natural spine, and switching would have introduced the inconsistency the finding was trying to prevent. Leaving as is.

**`L-006` pre-implementation enforcement.** Agreed and already handled in a2a4eba: the binding is now the rule ("the next free code in the band") rather than the number, so a concurrent allocation resolves rather than collides, with the §3.6 invariant as the catch. Nothing further to do at filing time without putting implementation code in a docs-only PR.

Noting the convergence: this round confirmed the `W-001` classification, the `amends`/033 relationship, and the `extends`-vs-`depends_on` pattern as non-findings, which matches the analysis in the previous replies. Those three are settled.

Gate unchanged on `a2a4eba`: `compile --check` 0, `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0, `cargo test --workspace` 0.

PR #91, opened 2026-09-06
feat(041-042): file wave 2 of the agentic-builder substrate

bartekus:
## Summary

Files wave 2 of `docs/design/02-agentic-builder-substrate.md`: spec 041 (the completion gate) and spec 042 (per-spec attestation). Both are `draft` / `implementation: pending`; no engine code changes.

**Both halves came out narrower than the design note sketched, because I checked the note against the code instead of implementing from it.** The note is updated to record the revision rather than quietly rewritten, so the original G1/G2 analysis stands with the corrections beside it.

## 041: `implementation: complete` defeats draft leniency

The note claimed `implementation: complete` was unverified self-assertion. That was over-stated. `index.rs:77` is:

```rust
fn in_flight(&self) -> bool {
    self.status == "draft" || matches!(self.implementation, Some(Implementation::Pending))
}
```

and spec 025 makes an unresolved owning unit a blocking `I-0xx` for any spec that is not in flight. So an **approved** spec marked `complete` is already held to its claims today.

What is actually missing is that `Implementation::Complete` never enters the expression, so `status: draft` alone decides. And that window is this repo's normal one, not a corner: a spec is filed as `draft` + `implementation: complete` in the PR that lands its code, and ratified to `approved` later. Spec 036 did exactly that at `46474a3`, ratified at `f2388ed`; spec 040 is in that state right now. So every spec passes through an interval where its strongest claim about code is the one claim nothing checks, and it is the interval when the claim is newest and least reviewed.

The fix is one arm of one predicate: a spec asserting completion is never in flight, whatever its `status`. §3.2 argues why the completion claim wins (it is the field *about* the condition being tested; `status` is about review, not the filesystem).

§3.3 is deliberately explicit about the ceiling: this verifies **existence**, not behavior. After it, `complete` means the claimed territory resolves. It does not mean the code is correct, and no mechanical gate here can.

**Blast radius: nil.** All 37 approved specs are already settled; 037-039 are `draft` + `pending` and stay lenient (which keeps 037's claim on the unwritten `verdict.rs` a warning); 040 is `draft` + `complete` and its section unit resolves. The committed index is unchanged.

It `amends` 025 and does not edit it, per spec 040 which landed in #90.

## 042: `attest --spec <id>`

A `SpecAttestation` scoped to one spec: `specSourceHash`, each owning unit with its resolved `contentHash`, and verdicts (`compile`, `resolution`, `lint`) restricted to that spec. It reuses 023's split exactly, pure payload with the clock and signer in the detached `LedgerSeal`.

**It is not committed, and that is the finding.** The note wanted a committed per-spec bundle for "build history in the repo as committed truth". `.gitignore:30` already keeps `.derived/attestation/` out of the tree because 023's attestation is on-demand by design, and committing a per-spec one would:

- restale on *every* edit to *any* claimed unit, since the payload hashes unit contents;
- need its own freshness gate, which is the note's own constraint (a fourth committed tree, a fifth gate verb, the lesson spec 031 exists to record);
- prove little that the working tree does not already show, since the signature is the part a compliance consumer wants and it is worth no more for sitting in git.

**§3.4 is normative and load-bearing:** no lint rule may consume a `SpecAttestation`, because the payload records the lint verdict and a lint conditioned on it would grade its own output. This is why 041 holds the completion claim with a severity tier rather than with evidence, and why the note's "lint requires the bundle before `complete`" pairing could not have worked as written.

## Testing

```
compile                            0   43 specs, 0 warnings
compile --check                    0   fresh, 43 shards
index coverage --fail-on-untraced  0   65/65 (100.0%)
index orphans                      0   none
couple --paths-from <diff>         0   2 paths checked, no drift
cargo test --workspace --locked    0   287 passed, 0 failed
```

## Reviewer notes

Two things previous rounds on #89 raised repeatedly, settled here in advance:

- `diagnostics` is **optional**: `codebase-index-spec-shard.schema.json:7` requires `["schemaVersion", "shardHash", "mapping"]`. Shards omit it when a spec has no diagnostics.
- 041 `amends` 025 **without editing** `specs/025-*/spec.md`. That is now written down in the corpus as of spec 040 (PR #90), in `standards/spec/contract.md` under "Amendment authoring". The inbound view is `spec-spine registry relationships 025-unresolved-unit-severity`.


bartekus:
Both findings valid and fixed. Verifying the first turned up an inaccuracy of my own, also fixed.

## Fixed

**`resolution.ok` conflated the fact with the severity tier (042 §3.1).** Correct, and the sentence was actively misleading. Spec 041 makes an unresolved owning unit blocking only for `implementation: complete`; an in-flight spec keeps `W-001`. Reading my cross-reference literally, an implementer could suppress `resolution.ok: false` for in-flight specs and produce an attestation asserting "resolution ok" for a unit that does not exist, which is precisely what the payload exists to make impossible.

Now stated as the fact: true iff every owning unit resolves, regardless of how the indexer tiers the failure. It also contradicted this spec's own §3.7 test ("an unresolved owning unit yields `resolution.ok: false` and still produces a payload"), so the prose was the wrong half.

**`approved` + `pending` was unnamed (041 §4).** Fair, and now named rather than left for a reader to notice. The combination is deliberate: `pending` is an honest statement that work has not been done, and this corpus files specs that are ratified before they are built, so the asymmetry with §3.1 is the point (`complete` and `pending` are opposite claims, and only one is falsifiable against the filesystem).

Your consumer concern is the part worth writing down, and §4 now does: `status: approved` says nothing about code, so anything reading the corpus programmatically MUST read `implementation`, and a consumer treating `status` as completeness will be wrong about every approved-and-unbuilt spec with no gate to contradict it.

**My own error, found while checking the above (041 §3.4).** I had written "every one of the 37 approved specs is `implementation: complete`". Not true: 36 are, and `000-spec-spine-bootstrap` carries no `implementation` key at all. The conclusion survives (an absent key is not `Pending`, so an approved spec without one is settled and its tier does not move) but the sentence did not, and a blast-radius claim is the wrong place to be approximately right. Corrected.

## Answered

**Ownership of `docs/design/02-agentic-builder-substrate.md`.** Confirmed rather than assumed, since you correctly noted it is not visible from the diff:

- Nothing owns it. Scanning every spec's owning edges (`establishes`, `extends`, `refines`, `constrains`, `co_authority`) across the whole corpus returns no claim on it, and the same holds for `docs/design/00-architecture.md` and `01-partial-supersession.md`.
- It could not reach the ownership check anyway: `docs/` is in the built-in bypass floor at `crates/spec-spine-core/src/couple.rs:38`.
- Empirically, `couple --base origin/main --head HEAD` exits 0 on this branch, and `self-governance` passed on this PR, which runs the same gate.

The `references`-only, `ownership: false` records in the 041 and 042 shards are the correct and intended shape since spec 034.

Gate after the fixes: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Both findings valid and fixed.

## Fixed

**`approved` + `in-progress` was absent from the §3.1 table.** You are right on the mechanism, and I modelled the predicate across all twelve combinations to be sure: `in_flight = status == "draft" || implementation == Pending` makes `approved` + `in-progress` **not** in flight, so a ratified spec whose implementation is openly underway gets hard `I-0xx` errors for units it has not written yet, both today and under my change.

Taking your first option. The table is now exhaustive over both enums, with only the `draft` + `complete` row moving and every other cell recording what the predicate already does, so an implementer cannot read a gap as a licence to guess.

I did not take the second option (extending the predicate to cover `in-progress`), and §4 now says why at length rather than leaving it implied. Briefly: it does look like a defect in 025 arm 2, since `pending` and `in-progress` are both statements that the work is unfinished and only one buys leniency, and 025 offers no argument for the omission. It bites an adopter who ratifies a design before building it, which is the same class of adopter-flow problem the OAP dry run raised against 025 originally; it does not bite this repo, where nothing is `approved` + `in-progress` because the local flow files `draft` + `pending`, builds, then ratifies.

But this spec argues exactly one thing, that a completion claim defeats leniency, and widening leniency for `in-progress` runs the opposite direction. Bundling them would produce a single spec that both tightens and loosens the same predicate for unrelated reasons. It deserves its own spec and its own argument, and §3.1 now states the current behavior precisely enough for that spec to amend.

**`resolution.ok: false` and the payload contract (042 §3.1).** Fair: §3.7 tested it and the normative text did not say it. Now stated where the payload is defined: a failing verdict never suppresses the attestation. `attest --spec <id>` emits a complete, hashable, signable payload whether verdicts are true or false, and exits `0` for having produced one. It is a record, not a gate, and evidence that refused to exist when the news was bad would be worth nothing.

Gate after the fixes: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple --base origin/main --head HEAD` 0.

bartekus:
All three valid and fixed. The second one improved the payload.

## Fixed

**Exit code as a verdict signal (§3.1).** Real footgun and now pinned normatively: `attest`'s exit code MUST NOT be read as a verdict, `0` means a payload was written and nothing about what it says, and `attest --spec X && next_step` is named as a misuse. It is the only verb in the tool where that holds, which is exactly why it needed stating instead of discovering, and the spec now points a caller who wants a gate at the verbs that are one.

Worth recording that this is **inherited, not introduced**: I checked `cmd_attest.rs`, and `attest` has always written its payload and returned `0` irrespective of the verdicts inside it. Spec 023 never documented that; `grep -n "exit" specs/023-ledger-seal/spec.md` returns nothing. So 042 now writes down what 023 left implicit, for both scopes, rather than creating an asymmetry between them.

I considered and rejected `--fail-on-false`, with the reasoning in the spec: a record verb that can be configured to refuse invites being used as a gate, which puts a second, weaker copy of the gate chain behind a flag when the verbs that already refuse are right there.

**`resolution.ok` uninterpretable without lifecycle (§3.1).** The best of the three, and it changed the payload. A phantom unit in a spec that is openly being built and a phantom unit in a spec asserting it is finished are the same `false` with very different meanings, and the payload gave a consumer no way to tell them apart.

Added `lifecycle: { status, implementation }`, with `implementation` omitted when the key is absent (as `000-spec-spine-bootstrap` has it). I took your option (a) rather than a `resolution.inFlight` flag deliberately: in-flight-ness is the indexer's severity policy, which spec 041 is changing in this very PR, so freezing a derived tier into an archival record would date it. `status` and `implementation` are the author's own words, which is what a record should keep. A test covers it.

**§3.4 is not mechanically enforceable.** Correct, and now admitted in the text rather than implied: nothing stops a contributor adding such a lint rule and passing CI, because "this rule reads an attestation" is not a property the build can test for. It is written in the spec that owns the payload so the argument is available at review time to whoever proposes one. Spec 040 §4 records the same honesty about its own rule, for the same reason: a convention that pretends to be enforced is worse than one that admits what it is.

## Answered

**Design-note ownership.** Confirmed for the third time, by construction rather than assumption: scanning every owning edge (`establishes`, `extends`, `refines`, `constrains`, `co_authority`) in all 43 specs returns no claim on `docs/design/02-agentic-builder-substrate.md`, and the same holds for `00-architecture.md` and `01-partial-supersession.md`. It could not reach the ownership check regardless, since `docs/` is in the built-in bypass floor at `crates/spec-spine-core/src/couple.rs:38`. `couple --base origin/main --head HEAD` exits 0 on this branch and `self-governance` has passed on every head of it.

**The `approved` + `in-progress` follow-on.** Agreed that the fix spec will amend 041, since §3.1 is now the authoritative table. It is recorded in 041 §4 with its reasoning, which is the tracker this project uses.

Gate after the fixes: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

PR #92, opened 2026-09-06
docs(042): pin attest's exit-code contract and make resolution.ok readable

bartekus:
## Summary

Carries onto `main` the three review fixes that PR #91 merged without. They were pushed to that branch after it had already been merged at `90a47cc`, so `371c278` does not contain them. This is a clean cherry-pick of `5c5b66e`: 39 added lines in `specs/042-per-spec-attestation/spec.md` plus the two regenerated shard hashes. No new authoring.

All three came from the final review round on #91, and all three were valid.

## What was dropped

**1. `attest`'s exit code was not pinned (§3.1).** Real footgun: every other verb puts its verdict in the exit code, so `attest --spec X && next_step` silently proceeds through a failed resolution. Now normative: the exit code MUST NOT be read as a verdict, `0` means a payload was written and nothing about what it says, and callers wanting a gate are pointed at the verbs that are one.

This is **inherited, not introduced**. `cmd_attest.rs` has always written its payload and returned `0` irrespective of the verdicts inside it, and `grep -n "exit" specs/023-ledger-seal/spec.md` returns nothing, so spec 023 never documented it. 042 now writes down what 023 left implicit, for both scopes, rather than creating an asymmetry between them.

A `--fail-on-false` flag was considered and rejected in the text: a record verb that can be configured to refuse invites being used as a gate, which puts a second, weaker copy of the gate chain behind a flag.

**2. `resolution.ok` was uninterpretable (§3.1).** The payload's most substantive gap. A phantom unit in a spec that is openly being built and a phantom unit in a spec asserting completion are the same `false` with very different meanings, and a consumer had no way to tell them apart.

Adds `lifecycle: { status, implementation }`, with `implementation` omitted when the key is absent (as `000-spec-spine-bootstrap` has it). Deliberately the author's declared lifecycle rather than a `resolution.inFlight` flag: in-flight-ness is the indexer's severity policy, which spec 041 changes in this same wave, so freezing a derived tier into an archival record would date it. A test covers it.

**3. §3.4's prohibition is a convention, not a mechanism.** Nothing stops a contributor adding a lint rule that consumes a `SpecAttestation` and passing CI, because "this rule reads an attestation" is not a property the build can test for. Now admitted in the text so the argument is available at review time to whoever proposes one, matching what spec 040 §4 records about its own rule.

## Testing

```
compile --check                    0   fresh, 43 shards
index coverage --fail-on-untraced  0   65/65 (100.0%)
couple --base origin/main --head HEAD   0   1 path checked, no drift
```

The regenerated `042` shard hashes are in the commit, so `compile --check` and `index check` are clean against it rather than merely against the spec text.


bartekus:
One valid, one that mistakes this repo's spec-first flow for a gate violation.

## Fixed

**Elliptical clause in the `lifecycle` bullet.** Right, and normative prose should not need its meaning recovered from the test bullet below it. "with `implementation` absent when the key is" is now "with `implementation` omitted when that key is absent from the spec's frontmatter."

## Not a violation

**"`lifecycle` is specified but the implementation is not updated."** Nothing in spec 042 is implemented, and that is the intended state, not an omission.

Spec 042 is `status: draft` / `implementation: pending`. The entire feature it describes is unbuilt: `attest --spec <id>` does not exist, nor does the `SpecAttestation` type, nor `SPEC_ATTESTATION_SCHEMA_VERSION`. `lifecycle` is not a new field being added to a shipped struct; it is one line of a payload that has never been emitted. You correctly noted you could not rule out that it was already implemented from REPO CONTEXT alone, so: it is not, and neither is anything else in the spec.

That is this repo's spec-first flow rather than an exception to it. A spec is filed as draft with its territory declared, and the code follows in a later PR; spec 025's severity tiers exist precisely so a pending spec can claim units that do not exist yet without failing the build (`W-001` rather than a blocking `I-0xx`). Specs 037 through 039 landed the same way in #89, and 037 claims `crates/spec-spine-types/src/verdict.rs`, a file that has never been written.

So no waiver is needed, and the gate agrees rather than merely tolerating it: `self-governance (compile · index · lint · couple)` passes on this PR, and `couple --base origin/main --head HEAD` exits 0 locally. A waiver would be the wrong instrument here regardless, since it records an accepted drift, and there is no drift to accept.

Worth adding for context: this PR is not new authoring at all. It is a cherry-pick of `5c5b66e`, three review fixes that were pushed to #91 after that PR had already been merged at `90a47cc`, so `371c278` on main does not contain them. The content was reviewed on #91; this only gets it onto main.

Gate on `27a81ce` plus the copy-edit: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
The second finding is right and is fixed. The first is the same one from the previous round, answered there.

## Fixed: the cross-spec normative claim

Good catch, and I had missed it. §3.1 states "the exit code of `attest` MUST NOT be read as a verdict", and that covers the **corpus-scoped** verb, which spec 023 owns and for which 023 carried no rule at all. Adding a normative rule to another spec's verb is an amendment, and 042 declared none.

Now declared: `amends: ["023-ledger-seal"]`, with the reasoning in §3.1. The alternative was to scope the MUST NOT to `--spec` alone and leave the corpus verb undocumented, which would put two different exit-code contracts on one binary for no reason beyond which spec happened to notice first.

Per spec 040 (merged in #90) the edge is declared in the amending spec and `specs/023-ledger-seal/spec.md` is not edited. The inbound view is live:

```
$ spec-spine registry relationships 023-ledger-seal
023-ledger-seal
  depends_on: 001-compile-registry, 003-conformance-lint, 004-codebase-index, 005-coupling-gate
  amended_by (incoming): 042-per-spec-attestation
  depended_on_by (incoming): 037-machine-readable-verdicts, 042-per-spec-attestation
```

This is the third amendment declared this way (041 amends 025, 038 amends 033) and the first one caught by review rather than authored deliberately, which is a fair argument that 040 was worth writing down.

## Already answered: `lifecycle` without an implementation counterpart

Same finding as the previous round, and the dichotomy it offers has a third answer that is the actual one: **nothing in spec 042 is implemented, and that is the intended state.** The spec is `status: draft` / `implementation: pending`. `attest --spec <id>` does not exist, the `SpecAttestation` type does not exist, `SPEC_ATTESTATION_SCHEMA_VERSION` does not exist. `lifecycle` is one line of a payload that has never been emitted.

"The test vectors and normative claims are wrong today" is not the right reading of a pending spec: they describe what the implementation must satisfy when it is written, which is what a spec is for in a spec-first corpus. Spec 025's severity tiers exist precisely so a pending spec can claim not-yet-written territory without failing the build, and spec 037 (merged in #89) claims `crates/spec-spine-types/src/verdict.rs`, a file nobody has written.

The gate agrees rather than tolerating it: `self-governance (compile · index · lint · couple)` passes on this PR, and `couple --base origin/main --head HEAD` exits 0 locally. No waiver applies, because a waiver records an accepted drift and there is none.

Gate after the amendment: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
Adopted the test suggestion. The `lifecycle` finding is closed; this is its third raising and it has been answered twice on this PR with evidence.

## Fixed

**Test coverage for the exit path §3.1 now mandates.** Good point, and it follows from the amendment rather than being incidental to it: 042 now carries a MUST NOT that reaches the corpus-scoped verb, so the exit path deserves a test that exercises it rather than an assumption that it still behaves as it did when nobody had written the rule down. §3.7 now requires both scopes to be covered: a corpus with a failing lint verdict, and a spec with `resolution.ok: false`, each write their attestation and return `0`.

Thanks for confirming the `amends: ["023-ledger-seal"]` reasoning. Worth noting it was your catch, and it is the third amendment declared this way (038 amends 033, 041 amends 025) but the first found by review rather than authored deliberately, which is a fair argument that spec 040 earned its place.

## Closed

**`lifecycle` has no implementation.** Third raising, answered twice above with the same evidence, which has not changed:

- Spec 042 is `status: draft` / `implementation: pending`. **Nothing in it is implemented** and nothing is meant to be by this PR: `attest --spec <id>`, the `SpecAttestation` type, and `SPEC_ATTESTATION_SCHEMA_VERSION` do not exist. `lifecycle` is one line of a payload that has never been emitted, not a field added to a shipped struct.
- "The spec is advancing ahead of the code" is a description of spec-first development, which is this corpus's constitution III, not a violation of it. Spec 025's severity tiers exist so that a pending spec can claim not-yet-written territory without failing the build; spec 037, merged in #89, claims `crates/spec-spine-types/src/verdict.rs`, a file nobody has written.
- The gate is not merely silent, it passes: `self-governance (compile · index · lint · couple)` is green on every head of this PR, and `couple --base origin/main --head HEAD` exits 0 locally.
- No waiver applies. A `Spec-Drift-Waiver:` records an accepted drift between code and its owning spec; there is no drift here to accept, and adding one would assert a violation that did not occur.

I will not re-answer this one again. If a future round raises it a fourth time, this reply is the answer.

Gate on `55d7fe9`: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
The `n-a` finding is valid and fixed in dfb6507. The `lifecycle` item is on its fourth raising and my previous reply stands as the answer.

## Fixed

**The `n-a` sentinel was unaddressed (§3.1).** Good catch. The bullet said `implementation` is omitted when the key is absent and said nothing about `n-a`, so an implementer could reasonably read it as another flavour of "not applicable" and suppress it. That would produce exactly the structural inconsistency you describe for a consumer comparing lifecycle fields.

Now pinned: `implementation` is omitted **only** when the key is absent, and every declared value is carried, `n-a` included. The two are different facts and the distinction is worth keeping: an absent key means nobody stated an intention, `n-a` means someone stated that none applies, and collapsing the second into the first loses the more informative of the two.

Also nailed down the spelling, since you raised the alias: `n-a` is emitted canonically, because spec 015 accepts `n/a` on the way in and normalizes it on the way out (`Implementation::Na` is `#[serde(rename = "n-a", alias = "n/a")]`). So a spec written in the predecessor dialect attests byte-identically to one written canonically, which matters for a payload whose whole value is that two people hashing the same corpus get the same bytes. A test covers both the presence of `n-a` and the dialect equivalence.

Thanks also for flagging that reviewers of 023 should know its exit-code contract is now governed from 042. That is the intended consequence of the amendment, and `registry relationships 023-ledger-seal` is where a reader of 023 finds it.

## Closed

**`lifecycle` has no implementation change.** Fourth raising. I said last round I would not answer it again, and I am holding to that: the reply two comments up is the answer. In one line, for anyone arriving fresh: spec 042 is `draft` / `pending`, nothing in it is implemented, and "the spec is ahead of the code" is a description of spec-first development rather than a violation of it.

Gate on `dfb6507`: `compile --check` 0 (43 shards), `index check` 0, `index coverage --fail-on-untraced` 0 (65/65), `lint --fail-on-warn` 0, `couple` 0.

bartekus:
No changes this round. One item is a settled question, the other matches nine existing specs.

**`depends_on` and `amends` both carrying `023-ledger-seal`.** Not redundant, and it is the corpus's universal pattern rather than a slip. Nine other specs do exactly this, seven of them approved: 009 (amends+depends_on 005), 025 (004), 026 (004), 027 (004), 030 (004), 034 (004), 036 (005), 038 (033), 041 (025).

They are different edges answering different questions, which is the same distinction spec 038 §4 argues for `extends` versus `depends_on`: `amends` says "I patch this spec's contract", `depends_on` says "this must come first". A spec can amend something it does not depend on, and depend on something it does not amend, so neither implies the other and the compiler is right not to infer one from the other.

Nothing stumbles on it. The two are separate fields in separate arrays, and the projection reads correctly in both directions:

```
$ spec-spine registry relationships 023-ledger-seal
  amended_by (incoming): 042-per-spec-attestation
  depended_on_by (incoming): 037-machine-readable-verdicts, 042-per-spec-attestation
```

042 appearing in both lists is the accurate answer, not a duplicate.

**`lifecycle` has no code counterpart.** Fifth raising. Closed, per my reply three comments up, which stands as the answer.

Thanks for confirming the amendment structure again. With no actionable findings this round, I am treating the review as converged and merging: all nine checks are green on `dfb6507`, and the substantive findings from the earlier rounds (the `amends` declaration you caught, the `n-a` sentinel, the exit-path test) are all in.

PR #106, opened 2026-09-06
ci(ai-review): count and review the diff without .derived/ shards; cap 2000 -> 3000

bartekus:
## What

The oversized-diff skip in `ai-pr-review.yml` counted regenerated compiler output. #104 was 2156 lines, 793 of them `.derived/` shards that `compile --check` and `index check` already verify byte for byte, so the AI review skipped on lines no reviewer should read.

- Exclude `.derived/**` from both the size count and the diff piped to the reviewer (new `REVIEW_DIFF_EXCLUDE` env, a git pathspec). #104 becomes 1801 reviewable lines.
- Raise `DIFF_SIZE_CAP` from 2000 to 3000, the upper end of a spec-plus-code PR in this corpus. The cap is a precision guard for a diff-only reviewer, not a context limit, so this stays conservative.
- The skip notice says the count is of the reviewable diff.

## Testing

- Same `git diff --numstat ... -- . ':(exclude).derived/**'` shape run locally over #104's range: 1801 lines, zero `.derived/` hunks in the excluded diff.
- Workflow YAML parses. `.github/` is on the coupling bypass floor; `couple` reports no drift.


bartekus:
Re the `**` finding: checked empirically over #104's range (25 `.derived/` files, nested two directories deep). All three forms exclude every one of them:

```
:(exclude).derived/**       -> .derived files remaining: 0
:(exclude,glob).derived/**  -> .derived files remaining: 0
:(exclude).derived          -> .derived files remaining: 0
```

Git's default pathspec match is fnmatch *without* `FNM_PATHNAME`, so a bare `*` does cross `/`; it is `:(glob)` that adds the pathname restriction, not the other way round. The original form was correct. Switched to the bare prefix `:(exclude).derived` anyway, since it says the same thing without depending on that fact.

PR #110, opened 2026-09-07
feat(049): `spec-spine verify <id>`, a spec's declared acceptance as a verb

bartekus:
Files and implements **spec 049**, `spec-spine verify <id>`. Audit item §3.1, named by spec 043 §4 as "the most substantial thing the adoption invented and the one most worth having" and deferred there; deferred again by spec 048 §4.

## Why now

Three adopters wrote the same 78-line `scripts/verify-spec.sh` independently, and #107 vendored a fourth copy into `kit/`. The kit currently ships the workaround this verb exists to retire.

## The seam

The engine never spawns a process. `core::verify::plan` is a pure function of the spec's markdown returning a `VerifyPlan`; `cmd_verify.rs` executes it. That is the same split spec 005 draws for `git`, and it is what makes the grammar testable over strings: the **26 fixtures in `core/tests/verify.rs` are the first assertions this parse has ever had** in any of the four repositories carrying it.

## Three deliberate divergences from the ported script

Each keeps an older contract intact, and none loses information:

| script | this verb | why |
|---|---|---|
| missing spec exits **2** | exits **1** | 2 means *stale* everywhere else in this tool |
| propagates the command's exit code (**137** on a signal) | exits **1**, code in `failure.exitCode` | the documented set is `0`/`1`/`2`/`3`; same defect class spec 035 fixed for a pipe |
| `not-declared` only in prose | `declared: false` in the payload | absence must be distinguishable from a pass without matching a sentence |

## Two things building it found

Both are recorded as dated decisions in the spec, not silent additions:

- **§3.7, `R-001`.** The first draft carried `spec-spine verify 049` inside 049's own `## Verification` block, which is the natural way to assert the envelope names this verb. Running it **forked 350 processes** before it was killed by hand. Re-entry is now refused, with the in-progress ids carried to children via `SPEC_SPINE_VERIFY_STACK` (the mechanism `couple` already uses for `SPEC_SPINE_PR_BODY`). Only a *cycle* is refused; any depth repeating no id still runs. The spec's block now demonstrates the refusal on itself.
- **§3.8, `--plan`.** Prints what would run without running it. For the one verb that executes what the corpus declares, reading the plan first is a safety affordance, not a convenience, and it is the natural CLI surface for the pure function the engine keeps.

## Schema

`VERDICT_SCHEMA_VERSION` **0.1.0 -> 0.2.0**. A new `verb` token is additive by the MINOR rule `version.rs` already states for `error.kind`. Nothing committed carries this version, so no shard changed shape. The spec 037 envelope test now tracks the constant rather than a literal, so a future additive verb does not read as a change to the existing six.

## Scope held back deliberately

Retiring `scripts/verify-spec.sh` and its kit copy is **out of scope** (§4) and is a follow-on. Removing the script from `kit/` would strand every adopter whose pinned `spec-spine` predates the release carrying this verb, and all four run two to four releases behind today. The retirement needs an `amends` edge on spec 048, which established the script; nothing here edits 048.

## Gate

| check | result |
|---|---|
| `compile --check` / `index check` | fresh, 50 shards |
| `lint --fail-on-warn` | 0 / 0 / 0 |
| `index coverage --fail-on-untraced` | 72/72, exit 0 |
| `couple --base origin/main --head HEAD` | **OK, 12 paths, no drift, no waiver** |
| `cargo test --workspace --locked` | green, 26 suites |
| `cargo clippy -D warnings` / `cargo fmt --check` | green |
| `spec-spine verify 049` | **passed (8 commands)**, through the verb it specifies |

Corpus 50; 049 lands `draft` and is ratified separately, per `AGENTS.md`.

Independent of #109 (the v0.15.0 bump): disjoint shards, which is what spec 024 sharding is for.


bartekus:
Addressed the AI review.

**Verification block assumes a pre-built release binary: real, fixed in `dc1a3ce`.** `cargo test` builds debug artifacts only, so an orchestrator running this in a clean checkout of the merged sha would reach the first `target/release/spec-spine` call with nothing there. The block now opens with `cargo build --release --locked` and is self-contained. `spec-spine verify 049` passes 9 commands.

**`README.md` / `docs/api.md` missing from the spec's edges: false positive.** Both are on the coupling gate's built-in bypass floor (`couple.rs::DEFAULT_BYPASS_PREFIXES` lists `docs/` and `README.md`), so neither needs a declared edge and neither can raise `C-001`. The empirical check agrees: `couple --base origin/main --head HEAD` exits 0 (12 paths, no drift) on this branch, and the self-governance job passed in CI. No waiver is needed or present.

PR #112, opened 2026-09-07
feat(050): index diagnostics reach a gate

bartekus:
Files and implements **spec 050**. Audit item §3.2, and the thing spec 048 deferred that `/burndown` needs.

## The gap, stated precisely

Spec 025 downgrades an unresolved unit to a counted `W-001` (owning unit on an in-flight spec) or `W-002` (non-owning `references` unit) rather than failing the index, because work under way legitimately claims territory that does not exist yet. **That is right and nothing here changes it.** What 025 never gave the warning is a reader.

The asymmetry is with the error tier, not with severity:

| surface | what it does with `W-001` / `W-002` |
|---|---|
| `index` (writing) | prints both tiers, then returns 0 |
| `index check` | ignores them; prints `index is fresh` |
| `lint --fail-on-warn` | counts `L-001`..`L-006`; never sees them |
| `index render` | shows them, inside markdown |

Error-tier diagnostics *are* gated (`BLOCKING_CODES` makes their shard read as stale, `index check` exits 2). Only the writing command reports warnings, and reporting is not its job: you have already changed the tree to find out. `cmd_index.rs` has carried the comment *"a warning the operator never sees is a unit that quietly went unresolved"* above that very loop. aicortex carries 248 that nothing reports.

## Three surfaces, none of which recomputes

- **`index check` reports the counts it already holds**, appended to the verdict line and added to the `--json` payload. A corpus with none keeps printing the bare `index is fresh`, so a clean tree reads exactly as before.
- **`index check --fail-on-unresolved`** turns them into exit 1. **Opt-in**, because specs 025 and 044 exist to let a ratify-then-build corpus stay green while work is openly under way; making it the default would refuse the workflow they were written to permit. **Staleness outranks it** (exit 2): a stale ledger's warnings describe a tree that no longer exists, so refusing for them would name the wrong problem.
- **`index diagnostics [--json]`** lists them, attributed to the spec whose shard carries them. `Diagnostic` has no spec member and `load_committed_index` flattens both shard sets, so attribution has to come from the shards rather than the aggregate.

All three read the **committed** shards. The counts describe the ledger the corpus actually compiled to, which is the only answer checkable against git.

## Design notes worth review

**`IndexCheckReport` lives in core** and is serialized by both `check_freshness_json` and the CLI's `--json` arm. The first attempt composed the payload twice and spec 037's facade-parity test caught it immediately, which is the test doing exactly its job. `compile --check` keeps the bare freshness object: index diagnostics are meaningless for the registry, and a permanently-zero member would be worse than none.

**The envelope version does not move.** Adding a member to one verb's `report` is additive, and `VERDICT_SCHEMA_VERSION` versions the envelope (`schemaVersion`, `verb`, `ok`, `exitCode`, `report`/`error`), not the per-verb payloads. §3.6 records the decision because 037's text is genuinely ambiguous and the next spec to add a payload field should not have to re-derive it.

## Two things building it turned up

- **The new gate immediately found a real `W-001` in this repo**: spec 050 claiming `tests/diagnostics.rs` before it was written. The in-flight leniency working as designed, and the first time this corpus could see it.
- **Audit item §3.9, reproduced.** `span_files_for_mapping` puts only `section`, `symbol` and `module` backing files in the shard hash, so a claimed plain `file` unit can be rewritten and `index check` still says fresh. Verified directly: rewriting a claimed `lib.rs` left the verdict at `is fresh`; editing the `spec.md` restaled it at once. Named in §4 because it **bounds what this gate promises** rather than being a stray observation: `--fail-on-unresolved` refuses a *recorded* unresolved unit, and recording happens when the index is regenerated. Fixing it changes spec 004's staleness contract and belongs in its own spec.

## Gate

| check | result |
|---|---|
| `compile --check` / `index check` | fresh, 51 shards |
| `index check --fail-on-unresolved` | **exit 0** (this corpus records none) |
| `lint --fail-on-warn` | 0 / 0 / 0 |
| `index coverage --fail-on-untraced` | 74/74, exit 0 |
| `couple --base origin/main --head HEAD` | **OK, 7 paths, no drift, no waiver** |
| `cargo test --workspace --locked` | green, 27 suites |
| `cargo clippy -D warnings` / `cargo fmt --check` | green |
| `spec-spine verify 050` | **passed (8 commands)** |

Corpus 51; 050 lands `draft` and is ratified separately. Stacks on #111 (the 049 ratify PR), which is independent.


bartekus:
Addressed the review. Three findings were real and are fixed in `471bfc0`; one is not a defect.

**1. Redundant guard around `counts_suffix`: real, and worse than described.** The split into `counts_summary` (bare) and `counts_suffix` (parenthesised) landed in the helpers but never reached that call site, so the stale line actually read `also records (1 warning(s), ...)` and the outer guard was load-bearing for a function that also self-guards. Now uses `counts_summary`.

**2. `count` silently promotes unknown severity: real.** `AttributedDiagnostic::severity` was a `String`, so "anything that is not `error` is a warning" was an invariant nothing enforced. It is now the shared `Severity` enum (serializes lowercase, so unchanged on the wire) and the fold matches exhaustively, with `Info` explicitly accounted for.

**3. `committed_counts` materializes the listing: real, fixed.** It now folds directly over the shards. The **second read of the shard set stays**, deliberately: folding it into `check_index_freshness` means refactoring the staleness gate that every other gate depends on, which is not worth a constant factor over small per-spec JSON files in a verb that already hashes every input. Recorded as a dated decision in §3.5 rather than left as an oversight.

**4. Prose example diverges from the implementation: real, and the example was the defect.** §3.1 printed `0 W-002` while stating two paragraphs below that zero entries are omitted. The example now matches what the code emits and says why.

**5. `lib.rs` should extend spec 004, not 001: not a defect.** An `extends` edge names the unit's *owner*, and `crates/spec-spine-core/src/lib.rs` is established by spec 001. `registry` confirms it, and spec 042 declares the identical edge for the same file when it added `attest_spec_json` to that same facade. Which verb the added function serves does not change who owns the file.

Gate re-run green: `couple` OK (7 paths, no drift, no waiver) · `compile --check` / `index check --fail-on-unresolved` exit 0 · coverage 74/74 · 27 test suites · clippy and fmt clean · `spec-spine verify 050` passed (8 commands).

bartekus:
Round two addressed in `fea12bd`. Both findings were fair.

**1. Ambiguous prose when the flag refuses: real, fixed.** stdout said `index is fresh (...)` while the process exited 1. Freshness is true on its own axis, so the line was not wrong, but a reader seeing only it would take it for a pass. It now reads:

```
index is fresh; refusing on unresolved units (1 warning(s), 0 error(s): 1 W-001)
```

and §3.2 states the rule normatively. The stderr line now names the flag and the counts instead of repeating the clause. Worth noting the scope: prose stays commentary, and both machine surfaces (exit code, `--json`) were already unambiguous, which is the whole point of spec 037. `index coverage --fail-on-untraced` has exactly the same shape today and is deliberately left alone; this is one clause, not a new contract for every gating flag.

**2. The read-cost note only covered the CLI: fair, corrected.** §3.5 now says the second shard read is equally `check_freshness_json`'s, and points a hot-path caller at the typed API, where `check_index_freshness` and `committed_counts` are separate calls it can order as it likes. That is the trade-off `check_registry_freshness_json` already documents against `compile_json`, so the facade's one-verdict-per-call contract stays consistent rather than acquiring an exception here.

Gate green: `couple` OK (7 paths, no drift, no waiver) · `compile --check` / `index check --fail-on-unresolved` / `lint --fail-on-warn` / `coverage --fail-on-untraced` all exit 0 · 27 suites · clippy and fmt clean · `spec-spine verify 050` passed (8 commands).

bartekus:
Round three addressed in `4e2cabf`. Both findings are corrections to my own round-two change, not new scope, so I fixed both rather than deferring.

**1. Double-print of the refusal: real, and self-inflicted.** Round two added the stdout clause and left the stderr line saying the same thing, so one fact went out twice on two streams. The stderr block is gone and the stdout arm names the flag:

```
index is fresh; --fail-on-unresolved refuses (1 warning(s), 0 error(s): 1 W-001)
```

stderr is now empty on that path. This matches `coverage --fail-on-untraced`, which reports on stdout and lets the exit code carry the refusal.

**2. §3.1 described the pre-`IndexCheckReport` structure: real.** The sentence claiming `compile --check` and `index check` "both render their payload through" `freshness_report` stopped being true when spec 037's parity test forced the composition into core. §3.1 now describes what shipped and records why: a payload built twice is a payload that drifts, and the parity test refused the first attempt correctly. The invariant it was protecting is unchanged and now stated directly, as `compile --check`'s payload must not change.

**On the observation that the tests asserted exit codes only:** that is exactly why the double-print survived a full CI run, so it is the more useful half of this review. The tests now assert the refusal appears once, on stdout, with stderr empty, and that no bare `index is fresh` line stands alone while the process exits 1.

Gate green: `couple` OK (7 paths, no drift, no waiver) · `compile --check` / `index check --fail-on-unresolved` / `lint --fail-on-warn` / `coverage --fail-on-untraced` exit 0 · 27 suites · clippy and fmt clean · `spec-spine verify 050` passed (8 commands).

bartekus:
Round four: **no actionable defect.** Checked each finding rather than assuming.

**Silent discard of package-shard diagnostics: false positive, verified.** `IndexPackageShard` has no `diagnostics` field. Its three members are `schemaVersion`, `shardHash` and `package` (`crates/spec-spine-types/src/codebase.rs:230`). There are no package-shard diagnostics to discard, which is also why `load_committed_index` accumulates diagnostics from the spec-shard loop only and its package loop pushes just packages and keyed hashes. Destructuring the package vector away is the correct read, not a dropped one. The reviewer was right that this could not be settled from the diff alone.

**`counts_summary` always emits both tier totals:** intentional, and the review says so. The zero *aggregate* is kept (`0 error(s)`) while the per-code breakdown omits zero entries; §3.1's example and the JSON agree.

**Exit code computed before the match, then re-guarded:** fair as a maintainability note. Left as is: `code` is derived once from the two facts that decide it, and the guard reads that single value rather than re-deriving it, so the two cannot disagree. A third code in the fresh case would be a spec change, not a refactor.

**`status: draft` + `implementation: complete`:** the normal flow in this repository, which files a spec as draft, builds it, then ratifies in a separate PR (`AGENTS.md`, "Working the backlog"). Spec 041 holds `complete` to its claims at any status, which is why the unresolved unit this spec introduced had to be written before the gate would pass, and spec 044 makes the in-flight combination explicit. 050 is ratified in its own follow-up PR.

Three rounds of real fixes landed (`471bfc0`, `fea12bd`, `4e2cabf`); this round found nothing to change. Merging.

PR #161, opened 2026-09-09
docs(079): a dead glob is dead in both tables

bartekus:
Files spec 079 as `draft`. No code changes; the build is a separate PR.

## The gap

`L-010` (spec 074 §3.1) refuses an `[index] extra_hashed_inputs` pattern ending in `/**`, because in the `glob` crate `dir/**` enumerates directories and the hasher keeps only files. It reads that one table.

`[index.slices]` (spec 012) holds pattern lists too. `IndexConfig::slices` documents them as carrying "`extra_hashed_inputs` pattern semantics", and the slice walk keeps only files exactly as the global one does. So `dir/**` is equally inert there, and nothing says so.

## Why this is not hypothetical

An adopter audited on 2026-09-09 carried **nine** dead patterns across the two tables. The six in `extra_hashed_inputs` were found and fixed; the two in `[index.slices]` survived the same fix pass **and a review**, because the fix was driven off the lint's output and nothing named them.

Both survivors pointed at directories that do not exist yet. That is the worse half: they cost nothing on the day they are written and stay silently empty on the day the directory arrives.

An adopter who upgrades is currently told about one table and not the other, which is worse than being told about neither, because the silence reads as clearance.

## The part that is not a copy of 074

§3.3. The current `L-010` message ends "so it can contribute no bytes to any content hash". That is true of `extra_hashed_inputs` and **false of a slice**: spec 012 keeps slices out of `contentHash` by design, and a *correct* slice pattern contributes no bytes to it either. Reusing the sentence verbatim would ship a false statement under a true code. The slice message must speak about the slice's own hash, the one `index check --slice <name>` gates.

## Verification fails against pre-079 code

The decisive line builds a fixture whose only defect is a dead pattern in a slice, and asserts `lint --fail-on-warn` exits 1. Today it exits **0**, confirmed:

```
lint: 0 error(s), 0 warning(s), 0 info
  lint --fail-on-warn exit: 0
```

The block also pins the boundary rather than the presence of a warning: the corrected `**/*` form must pass. This repository declares no `[index.slices]` table, so its own lint cannot demonstrate the rule and could not regress if the rule were deleted, which is why acceptance is on a fixture (§3.4).

## Two edges worth a look at review

**The lint unit is claimed through `003-conformance-lint`, not through 074**, even though 074 is where `L-010` came from. 074 itself carries `lint.rs` through 003, so this follows the path already taken for the file; and 074 is still `draft`, held only by its §3.6 about deleting `kit/scripts/verify-spec.sh`, so depending on it would report 079 blocked behind a decision it does not wait on. Recorded as D-1.

**One code, not `L-011`.** Same defect, same reasoning, same remedy; only the sentence about what is lost differs, which §3.3 handles. Recorded as D-2.

## Out of scope, deliberately

`specs/012-index-hash-slices/spec.md:62` teaches the dead form, `workflows = [".github/workflows/**"]`, and adopter slice tables were copied from it. 012 is approved, and a new spec does not get to rewrite an approved spec's text through its own decision entry. That correction is a direct one-token human edit, taken outside this spec.

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees, `couple` 0 with no waiver.

### DIFF
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "2eb5710ba4c613add93fa1dea847ebf791ff9389cf2d5591b7de014603ff407a"
+  "shardHash": "5df03f856f0ca46d554c57277af2ef9c6e5d59c4a476b29470c97f108624e64a"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "5e19cbc4e0d9789221ba38f7fbdae728bef034ed8eb80f638cc94068a215cec7"
+  "shardHash": "da221e294ef1cf14ce9f7b749557190537ff2776a353c156d44bbbc6e1ae7f74"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "5df03f856f0ca46d554c57277af2ef9c6e5d59c4a476b29470c97f108624e64a"
+  "shardHash": "5a9a6bab7f67f37752a68119f0b78e6ed2ef64e0478c4dbf66b41fc62643bb68"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "b02997bf5959d9c8b9a3457164e3b5c5710b9057616e50abab73a90bd853205f"
+  "shardHash": "fadcae0fa0e2e25f086893f19e2fdb36fe5663a860bc0eab720c3f0a68b3b496"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "da221e294ef1cf14ce9f7b749557190537ff2776a353c156d44bbbc6e1ae7f74"
+  "shardHash": "b02997bf5959d9c8b9a3457164e3b5c5710b9057616e50abab73a90bd853205f"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "fadcae0fa0e2e25f086893f19e2fdb36fe5663a860bc0eab720c3f0a68b3b496"
+  "shardHash": "af1347fe70448e10d1ed8f8cf61581b7537b689c4ef7eca381a12f87ff0105ed"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -1,0 +1,52 @@
+{
+  "mapping": {
+    "dependsOn": [
+      "003-conformance-lint",
+      "012-index-hash-slices",
+      "053-depends-on-ordinal-monotonicity",
+      "069-the-shipped-default-hashes-what-it-names"
+    ],
+    "implementingPaths": [
+      {
+        "path": "crates/spec-spine-core/src/lint.rs",
+        "source": "spec-edge"
+      },
+      {
+        "path": "crates/spec-spine-core/tests/lint.rs",
+        "source": "spec-edge"
+      }
+    ],
+    "resolvedUnits": [
+      {
+        "locations": [
+          {
+            "file": "crates/spec-spine-core/src/lint.rs"
+          }
+        ],
+        "ownership": true,
+        "sourceField": "extends",
+        "unit": {
+          "kind": "file",
+          "path": "crates/spec-spine-core/src/lint.rs"
+        }
+      },
+      {
+        "locations": [
+          {
+            "file": "crates/spec-spine-core/tests/lint.rs"
+          }
+        ],
+        "ownership": true,
+        "sourceField": "extends",
+        "unit": {
+          "kind": "file",
+          "path": "crates/spec-spine-core/tests/lint.rs"
+        }
+      }
+    ],
+    "specId": "079-a-dead-glob-is-dead-in-both-tables",
+    "specStatus": "draft"
+  },
+  "schemaVersion": "1.1.0",
+  "shardHash": "2eb5710ba4c613add93fa1dea847ebf791ff9389cf2d5591b7de014603ff407a"
+}
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "5a9a6bab7f67f37752a68119f0b78e6ed2ef64e0478c4dbf66b41fc62643bb68"
+  "shardHash": "5e19cbc4e0d9789221ba38f7fbdae728bef034ed8eb80f638cc94068a215cec7"
 }
--- a/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/codebase-index/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,5 +48,5 @@
     "specStatus": "draft"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "af1347fe70448e10d1ed8f8cf61581b7537b689c4ef7eca381a12f87ff0105ed"
+  "shardHash": "8576a124905c57b8fa8b07b91dfa2d9d9e903fb73d357b00a5b9a2c16a074ca7"
 }
--- a/.derived/spec-registry/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/spec-registry/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -48,6 +48,6 @@
     "summary": "`L-010` refuses a pattern ending in `/**` in `[index] extra_hashed_inputs`, where it enumerates directories and can contribute no bytes. `[index.slices]` carries pattern lists with the same documented semantics and the same file-only walk, and the lint is silent about it. An adopter audited on 2026-09-09 carried seventeen dead patterns, nine in the table the lint reads and eight in the table it does not. A fix pass corrected all nine and six of the eight; the two survivors reached an already-reviewed pull request and were caught by a second human reading, since no tool could report them. Both pointed at directories that do not exist yet, so they would have stayed silently empty on the day those directories arrived. This spec extends the rule to the second table, and requires the slice message to speak about the slice's own hash rather than repeating a sentence about `contentHash` that is false for slices.\n",
     "title": "A dead glob is dead in both tables"
   },
-  "shardHash": "5bcfe8d5bfaebf445c02d5226e40438fd1f4279ffac88dbb5e9b74ed970c772b",
+  "shardHash": "73393c1dfb78ee4b9f46b070c2090220cab3c8ed3be61242fa70635d9d0fdaa9",
   "specVersion": "1.2.0"
 }
--- a/.derived/spec-registry/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
+++ b/.derived/spec-registry/by-spec/079-a-dead-glob-is-dead-in-both-tables.json
@@ -1,0 +1,53 @@
+{
+  "record": {
+    "created": "2026-09-09",
+    "dependsOn": [
+      "003-conformance-lint",
+      "012-index-hash-slices",
+      "053-depends-on-ordinal-monotonicity",
+      "069-the-shipped-default-hashes-what-it-names"
+    ],
+    "extends": [
+      {
+        "nature": "additive",
+        "spec": "003-conformance-lint",
+        "unit": {
+          "kind": "file",
+          "path": "crates/spec-spine-core/src/lint.rs"
+        }
+      },
+      {
+        "nature": "additive",
+        "spec": "053-depends-on-ordinal-monotonicity",
+        "unit": {
+          "kind": "file",
+          "path": "crates/spec-spine-core/tests/lint.rs"
+        }
+      }
+    ],
+    "id": "079-a-dead-glob-is-dead-in-both-tables",
+    "implementation": "pending",
+    "kind": "tooling",
+    "owner": "The spec-spine Authors",
+    "risk": "low",
+    "sectionHeadings": [
+      "079: A dead glob is dead in both tables",
+      "1. Purpose",
+      "2. Territory",
+      "3. Behavior",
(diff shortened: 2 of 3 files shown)


bartekus:
Thanks. Three findings were real and are fixed in 9b6a5bd; two do not apply, with the evidence below.

## Fixed

**§3.3 double invocation and the `&&` guard.** Agreed the compound line was doing too much. The preceding line already asserts a slice line exists, so this one only has to be negative. Now one invocation:

```
! target/release/spec-spine --repo "${TMPDIR:-/tmp}/ss079" lint 2>&1 | grep 'index.slices' | grep -q 'content hash'
```

On the spurious-failure path you describe: a line containing both `index.slices` and `content hash` is exactly what §3.3 forbids, so catching it is the intent rather than a false positive. The rewrite keeps that and drops the redundant run.

**Unexplained `depends_on` on 069.** Correct, nothing in the body cited it. Rather than drop it, §1 now states the relationship, because it is load-bearing: 069 fixed the shipped default and, by its own §4, *deferred* this lint on the forward-looking-entry argument; 074 §3.1 took it up once the form was shown inert under every tree rather than merely empty under this one; neither pass looked at the second table.

**§4's "before it".** The strongest finding. That sentence asserted a sequencing that has not happened: `specs/012-index-hash-slices/spec.md:62` still reads `.github/workflows/**` on `main`. §4 now says so explicitly, and says that 079's acceptance does not depend on it and that 079 can be built and ratified with 012 unchanged. Tracked in the open rather than implied.

## Not changes

**`establishes` is a real field.** It is one of the eight typed edges of this system, declared in `crates/spec-spine-types/src/frontmatter.rs` at line 102 (known keys) and line 156 (`pub establishes: Vec<Unit>`). There is no `owns:` field to switch to. The fixture also parses in practice: running it today yields `lint: 0 error(s), 0 warning(s), 0 info`, which is the pre-079 pass the §3.1 assertion is designed to fail against, so it is not a false positive.

**`; test $? -eq 1` is safe here.** `verify` runs each fence line as its own process, `sh -c "<line>"`, with no `set -e` anywhere (`crates/spec-spine-cli/src/cmd_verify.rs:114-120`). Each line's status is captured and compared individually. That is why spec 059, approved and `implementation: complete`, already uses the same pattern.

## Re-verified after the edits

The block still fails against pre-079 code, which is the property that matters for a draft:

| assertion | exit today |
|---|---|
| slice line present in `lint` output | 1 |
| `lint --fail-on-warn` refuses the fixture | 1 |

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

bartekus:
Second round addressed in f75a3a4. The `tests/lint.rs` question settles cleanly, and the `contentHash` catch was a real hole in my acceptance.

## The `053` route is correct, and now says why

Settled from the corpus: **each unit is routed through the spec that `establishes` it.**

- `003-conformance-lint` establishes `crates/spec-spine-core/src/lint.rs` (`specs/003-conformance-lint/spec.md:12`).
- `053-depends-on-ordinal-monotonicity` establishes `crates/spec-spine-core/tests/lint.rs` (`specs/053-depends-on-ordinal-monotonicity/spec.md:29`), because it **created** that file: `L-007` was the first rule to need a dedicated lint test file.

So `053` does not merely co-own it, it owns it, and no coupling violation awaits the implementation PR.

Your quality point stands regardless and D-1 now covers both units under one rule. I also recorded something worth knowing that turned up while checking: **the corpus is not consistent here.** Specs 057 and 058 carry `tests/lint.rs` through `003`, which does not own it. Nothing refuses them, because an `extends`-carried unit is a first-class claim whether or not the target established it. That is exactly why stating the routing rule is worth the lines.

## The `contentHash` fragility was a real hole

Good catch, and the consequence was worse than fragility: the negative assertion would have passed **trivially** against an implementation emitting `contentHash`, leaving §3.3 untested while looking green.

Fixed on both sides rather than just the grep, so the spec binds the implementation instead of the test chasing it:

- §3.3 now prohibits the claim **in either spelling**, and says why the spelling matters.
- The assertion is `grep -qiE 'content ?hash'`, verified to match `content hash`, `contentHash` and `contenthash`.

## On the "user-visible regression" framing

Taking the substance, correcting the framing. `lint` reads `spec-spine.toml`, not spec markdown, so 012's fenced example changes no verdict for anyone and nothing regresses in the tool relative to before the lint existed. The cost is entirely on the reading side: an adopter copies the example and writes a pattern the tool then refuses, which is how at least one adopter's dead slice table got there. A documentation defect with a measured victim, not a behavior defect.

Your ask about ownership and timing is right and §4 now answers it: the edit is a human's, and it **should land before 079 is ratified**, so no window exists in which the corpus has ratified a rule its own specification contradicts. Building and merging 079 are fine with 012 unchanged. It is the ratifier's check rather than an acceptance criterion, since 079 cannot assert anything about a file it does not claim.

## Re-verified

| assertion | exit today |
|---|---|
| `lint --fail-on-warn` refuses the fixture | 1 |
| slice line present in output | 1 |

Still fails against pre-079 code, which is the property that matters for a draft. Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

bartekus:
Both findings were real. Fixed in ddbc326.

## The tier was half-proven, and that is the better catch

You are right, and the failure mode is the one worth guarding: §3.1 claims warning tier, the block asserted the `--fail-on-warn` refusal, and nothing asserted that a bare `lint` exits 0 on the same fixture. An implementation emitting `L-010` at **error** tier would have satisfied every other line while contradicting §3.1, and the acceptance would have looked green.

That is worth more than the added line, because promoting a warning is the natural reflex once a rule turns out to have no legitimate counter-example, which is exactly what §3.1 argues about this one. The tier is the claim most likely to be lost in implementation and was the claim least defended.

Fixed on both sides rather than just the block, so the requirement is stated rather than implied by a test:

- The block now runs bare `lint` on the tripping fixture.
- §3.4 now requires that assertion and says why, so the line is not an orphan a later edit could drop as redundant.

Note this line **passes today**, which is correct: it is a boundary control like the corrected-form check, not a fails-first assertion. The two lines that must fail against pre-079 code still do.

## Unescaped dot

Correct, and cheap. `grep 'index.slices'` read the dot as any-character, so the filter was wider than intended. Both occurrences are now `-F`.

## Failure profile after the changes

| assertion | exit today | role |
|---|---|---|
| `lint --fail-on-warn` refuses the fixture | 1 | must fail pre-079 |
| slice line present in output | 1 | must fail pre-079 |
| bare `lint` exits 0 on the fixture | 0 | boundary control (new) |
| corrected `**/*` form passes | 0 | boundary control |

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

## On your two design notes

Both read the spec the way it was written, and neither needs a change. The 012 example is a ratification gate rather than an acceptance criterion, and it is still outstanding on `main`. And building 079 without 074 complete is the intended consequence of D-1: `lint.rs` already carries the `L-010` path from 074's partial implementation (PR #150, ten of eleven sections), so the implementer is extending live code, not code that only exists in a draft.

bartekus:
Finding 1 is a real hole and is fixed in 8e94b58, though not the way you suggested, for a reason worth stating. Finding 2 does not apply. Finding 3 is recorded.

## 1. The multi-line hole is real

You are right that `grep -F 'index.slices' | grep -qiE 'content ?hash'` inspects only the naming line. If a message were split across a summary line and a continuation, the forbidden text could sit one line below the filter and the assertion would pass vacuously, defeating the section it exists to enforce.

**I did not take the `tee` fix, because an unscoped negative is unsound here.** `! grep -iE 'content ?hash' out.txt` would also match the `extra_hashed_inputs` form of `L-010`, which says "content hash" **legitimately and correctly**. That assertion would refuse a true message the moment the fixture grew a dead entry in the other table, and it would do so for the opposite of the right reason.

So I took your second option, which pins the invariant rather than working around its absence. §3.2 now requires **one line per violation**, as every existing `L-` code already emits, and states that this is what makes §3.3 checkable rather than a formatting preference. §3.4 now says the negative assertion MUST stay scoped to the slice's line, and why the unscoped form is wrong. The implementation is pinned; the acceptance no longer depends on an unstated assumption.

## 2. The fixture is valid, and last round's line already proves it

Three independent confirmations:

- **Empirically**: `lint` on that exact fixture exits **0** with `0 error(s), 0 warning(s), 0 info`. A config error would surface here.
- **By contract**: `standards/spec/contract.md:13` says `spec-spine.toml` has **all keys optional**, absent meaning conventional defaults. There is no `content_paths` or namespace to omit.
- **By precedent**: spec 059's fixture writes an **entirely empty** `spec-spine.toml` (`: > .../spec-spine.toml`) and is approved and `implementation: complete`.

Your concern about a wrong-reason pass is the right shape of worry, and the tier line added in the previous round already closes it: the block asserts bare `lint` exits **0** on this fixture. An invalid config cannot produce that, so a config failure can no longer be mistaken for the refusal.

## 3. `depends_on` as routing

Fair, and now written down. D-1 records that 053 appears in `depends_on` for ownership routing rather than behavioral precedence: 079 needs nothing 053 decided about ordinal monotonicity, it needs the file 053 created. Better in the spec than left for a reader auditing the graph to find surprising.

## Failure profile unchanged

| assertion | exit today | role |
|---|---|---|
| `lint --fail-on-warn` refuses the fixture | 1 | must fail pre-079 |
| slice line present in output | 1 | must fail pre-079 |
| bare `lint` exits 0 on the fixture | 0 | boundary control |
| corrected `**/*` form passes | 0 | boundary control |

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

bartekus:
Both findings are correct and fixed in f071d58. The first one lands squarely, and it is the third time this block has had the same defect.

## 1. The acceptance did not enforce the rule it declares

Last round I made §3.2's one-line rule load-bearing, on the grounds that §3.3's negative assertion selects the line naming the table. Then I left the positive side as three independent greps, which prove only that each string appears *somewhere*. An implementation splitting the message across lines would satisfy all three, leave §3.2 unverified, and render §3.3's scoped negative toothless.

That is the exact failure §3.2 was written to prevent, sitting unguarded in the block meant to prove it. Declaring an invariant and not testing it is worse than not declaring it, because the negative assertion now leans on an assumption the acceptance never checks.

The three greps are now one chain, which proves all three share a line:

```
target/release/spec-spine --repo "${TMPDIR:-/tmp}/ss079" lint 2>&1 | grep -F 'L-010' | grep -F 'index.slices' | grep -q 'workflows'
```

I used a single chain rather than your two-pipeline form because two pipelines prove `L-010` co-occurs with `index.slices` and that `index.slices` co-occurs with `workflows`, which with more than one violation could be satisfied by two different lines. The chain is strictly stronger and shorter. Verified to still fail against pre-079 code (exit 1).

## 2. D-1's lost paragraph break

Correct, and my doing: inserting the `depends_on` note last round left the argument for keeping 074 out of the edge lists trailing the routing paragraph, on an unwrapped 83-character line. D-1 is now three paragraphs, one point each: the routing rule and the corpus's inconsistency about it, why 074 is absent, and why 053 appears in `depends_on`.

## 3. Design note

No change. The fixture lints clean (`0 error(s), 0 warning(s), 0 info`), so the corrected-form assertion cannot fail for an unrelated code, and the bare-`lint` line added in round three already asserts exactly that: it requires exit 0 on the tripping fixture, which pins that nothing but the rule under test fires.

## Failure profile

| assertion | exit today | role |
|---|---|---|
| code, table and slice share one line | 1 | must fail pre-079 |
| `lint --fail-on-warn` refuses the fixture | 1 | must fail pre-079 |
| bare `lint` exits 0 on the fixture | 0 | boundary control |
| corrected `**/*` form passes | 0 | boundary control |

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

bartekus:
Confirmed and fixed in 3cfc438. You caught a factual error in the paragraph whose entire job is the evidence, and the arithmetic tell was there to see: "nine across the two tables" followed by "six" plus "two".

## Measured from the adopters' own history

| repo | `extra_hashed_inputs` | `[index.slices]` | total |
|---|---|---|---|
| aicortex | 9 | 8 | **17** |
| hqgit | 6 | no slices table | 6 |

The fix pass corrected all 9 in `extra_hashed_inputs` and 6 of the 8 in `[index.slices]`. The survivors were `deploy/**` and `eval/**`, both on multi-pattern lines. The "six" I attributed to `extra_hashed_inputs` was hqgit's number, and hqgit has no slices table, so the sentence blended two repositories into one adopter.

## The corrected version argues the point better

I had written that the slice patterns survived "because the fix was driven off the lint's output". The real split is sharper: **the table `L-010` reads came out of the pass clean, and the table it does not read came out of the same pass still broken, with nothing able to say so.** The two survivors were caught by a second human reading, not by any tool, because no tool can report them. That is the case for this spec stated in one sentence, and my miscount had blunted it.

The summary carried the same numbers and is corrected with it.

## On the category

You are right that this is the class I had named as the weak one, claims about what exists, and it landed in the worst possible paragraph. Worth noting the shape for anyone reading later: the prose numbers were never checked against the repositories, while every mechanical claim in the spec had been. The gate verifies the change; nothing verifies the description of the change, which this corpus has hit before.

I have recorded the per-table counts in memory with an explicit warning not to blend the two repositories.

## Sequencing, with your update

Both glob PRs merged (hqgit 5f63a4a, aicortex 350b79b), which I have confirmed. So: merge #161, build 079, and the two #6 pin PRs can rebase now and will move no index shards. Unchanged and yours: the one-token edit at `specs/012-index-hash-slices/spec.md:62` before 079 is ratified, and the amendments that unblock the pins (hqgit B-2 and B-3, aicortex 000 §9 and its contract).

bartekus:
One finding does not apply, one is fair as a documentation gap and is now closed, one is cosmetic and taken. 291de1e.

## `establishes` for the spec's own file: not required, and not the convention

Three independent confirmations:

- **Convention.** Only **2 of the 80** specs in this corpus claim their own `spec.md`. 076 and 077, the two most recent before this one, do not. 078 does, but for an unrelated reason: it establishes `AGENTS.md` and the governance surface around it.
- **Mechanism.** `coverage.rs::SOURCE_EXTS` is `rs, ts, tsx, js, jsx, go, py, sh`. **`md` is not in it**, so a spec document is never in the coverage universe and `index coverage --fail-on-untraced` cannot report it, nor can `C-002`, which classifies changed *source* files.
- **Empirically.** `compile --check`, `lint --fail-on-warn`, `check` and `couple` are all green on this branch, and CI's `self-governance` job passes. Your specific prediction, that `compile --check` in the fence would fail against this spec's own directory, does not happen.

The `establishes` in the fixture is there because the fixture's spec needs *some* territory for the corpus to be meaningful, not because a spec must claim itself.

## The `; test $? -eq 1` question is fair, and now answered in the spec

This is the second round to raise it, which tells me the block was relying on a harness behavior it never cited. That is a real documentation gap even though the code is fine.

Settled: each line runs as its own `sh -c`, a full POSIX shell, no `set -e`. The call is `Command::new("sh").arg("-c").arg(command)` in `crates/spec-spine-cli/src/cmd_verify.rs`, which executes each planned command and compares its status individually. So `$?` is the preceding command's status in the same shell and the assertion is real. Spec 059, approved and `implementation: complete`, relies on the same guarantee, which answers your "no other fence uses this pattern": one does, it is just not in this diff.

The preamble now states all of that, with your reasoning for why it matters: a reader assuming a restricted executor would read the tier assertion as vacuous, and the tier is the claim this spec most needs to hold.

## While correcting that, I found the preamble overstating its own profile

It said "every assertion below fails against pre-079 code". Two do: the co-occurrence chain and the `--fail-on-warn` refusal. The bare-`lint` tier line and the corrected-form line are **boundary controls** and pass on both sides, which is what makes them controls. The preamble now says which are which, so the ratifier is not told to expect four failures and find two.

## `-F` consistency

Taken, no argument. The third link is now `grep -qF 'workflows'`.

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees, `couple` 0 with no waiver.

bartekus:
Two real, one disproven. 0c0c991.

## 1. The refusal assertion did not name its own cause

Correct, and the regression case is sharper than the fixture case you raised: `lint --fail-on-warn ; test $? -eq 1` would keep passing if **this rule were later deleted** while some other warning took its place. It asserts "something warned", not "this rule fired".

Fixed by capturing the same invocation's output and requiring both conditions, on one line so no state crosses a line boundary:

```
o=$(target/release/spec-spine --repo "${TMPDIR:-/tmp}/ss079" lint --fail-on-warn 2>&1) ; test $? -eq 1 && printf '%s\n' "$o" | grep -F 'L-010' | grep -qF 'index.slices'
```

§3.4 now states the requirement, so the test is not the only thing holding it. Verified both ways: exit 1 against pre-079 code, and exit 0 against a simulated implementation that satisfies both halves.

## 2. `content-hash` would have slipped through

Correct and cheap. §3.3 said "whichever spelling" and named two forms; the regex matched exactly those two, so the section was broader than its own acceptance. §3.3 now names all three and the match is `content[ -]?hash`, verified against `content hash`, `content-hash`, `contentHash`, `Content Hash` and `CONTENT-HASH`.

That is the third time the gap between what a section says and what its assertion checks has been the finding, which is a useful pattern to have surfaced.

## 3. The fixture emits nothing

Testable, so I tested it rather than reasoning from the diff. That exact fixture:

```
lint: 0 error(s), 0 warning(s), 0 info
  bare lint exit: 0
```

`kind`, `owner`, `risk` and `implementation` are optional. Per `standards/spec/contract.md`, required frontmatter is `id`, `title`, `status`, `created` and `summary`, all of which the fixture has. So no error-tier diagnostic exists to break the bare-lint assertion, and the "only defect" claim holds as written.

Worth noting your finding 1 also hardens this: now that the refusal line requires the output to name the slice violation, an unrelated warning can no longer satisfy it even if one appeared later.

Gate: `compile` 80 specs 0 warnings, `lint` clean, `check` fresh on both trees.

PR #170, opened 2026-09-10
docs: CLAUDE.md names the ten skills the kit ships

bartekus:
## What

`CLAUDE.md` still said the kit ships "the fifteen skills". Spec 081 cut it to
ten on 2026-09-09 and updated `AGENTS.md`, `kit/AGENTS.md` and `kit/README.md`;
`CLAUDE.md` was not in 081's territory (§2 names the three listings, not this
file) and was left behind. `/prime` reads `CLAUDE.md` as identity, so every
session since has opened on a false count.

## Why not a spec

`spec-spine index owner CLAUDE.md` reports no owner, and `.md` is not in
`coverage.rs::SOURCE_EXTS`, so neither `C-001` nor `C-002` is in play. This is
unowned prose being corrected to match shipped behavior, not a change to what
anything requires. `couple` confirms: 1 path checked, no drift.

## Also corrected in the same pass

The file was audited line by line rather than patched at the one known error.
Three further claims were wrong or incomplete, each checked against the code:

| claim | was | is |
|---|---|---|
| `// Spec:` header window | "the first lines" | first **16** lines (`index.rs::scan_comment_headers`) |
| accepted markers | "`//`, not `//!`" | `//` **or `#`**, not `//!`. `SOURCE_EXTS` includes `py` and `sh`, where `//` is not a comment |
| skills parity pin | spec 048 | specs 048 **and 081** |

and `index owner` was added to the verb reference: it is the governed read for
"which spec owns this path" (spec 055), and it is what answered the ownership
question above.

## What was checked and left alone

The toolchain pin (1.92.0) and MSRV (1.85), `SOURCE_EXTS`, the four determinism
triples in `determinism.yml`, both lifecycle enums in `frontmatter.rs`, the
schema versions (registry 1.2.0, index 1.1.0, verdict 0.3.0), the
`required_version = ">=0.17.0"` floor, the core module list, and the "a quarter
of tracked source files have no establisher" figure, which still measures 20/79
with 16 of them in `spec-spine-types`.

## The shard churn

`CLAUDE.md` is an `[index] extra_hashed_inputs` member, so this edit restales
the whole index. The 86 regenerated shards (82 `by-spec`, 4 `by-package`) ship
with the change that made them stale. No registry shard moves: a registry shard
hashes `spec.md` source bytes, which this does not touch.

## Gate

`compile`, `index`, `check --fail-on-unresolved --fail-on-warn`,
`lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple`, then
[...]

PR #171, opened 2026-09-10
feat(082): a refusal is not a remediation round

bartekus:
Files and builds spec 082 (`status: draft`, `implementation: complete`;
ratify follows). This is the item spec 081 §4 deferred: the severity triage
the removed `validate-and-fix` carried, folded into `/shepherd` on its own
spec rather than smuggled in under a removal.

## The defect

`/shepherd` has one bound and one escalation. The bound is two remediation
rounds (048 §3.1). The escalation is the coherence guard, and it lives at
step 2 of the remediation procedure: it is read only once the session has
fetched the log, entered the fix, and is deciding how to edit. That ordering
is backwards for the one case it covers, and three further classes are not an
agent's to fix either, with the skill naming none of them:

- a hand-edit to an artefact the path-scoped rule names, the derived tree
  included;
- a dependency cycle (which edge to drop is a question about authority);
- an ambient input reaching a hashed path, including a determinism job red on
  some platforms and green on others.

Each presents as an ordinary red check. A session that classifies nothing
attempts one, spends a round, pushes, waits out a full 45-minute watch cycle,
fails again, spends the second round, and only then reports that a human was
always the answer, having by then put two commits on the branch that somebody
has to read. The bound protects the repo from an agent looping; it does not
protect the human's time, and it does not stop an agent acting inside a round
it was never entitled to spend.

## The fix (§3.1 to §3.4)

A new **Step 2: classify before you edit** runs off the run log the skill
already fetches; the old steps shift down by one. CRITICAL stops the skill and
**consumes no round**. The four CRITICAL rows are a closed list on purpose
(D-2): an open list is a judgement call at the moment an agent is most
motivated to judge generously.

HIGH, MEDIUM and LOW are the session's, fixed in that order within a round.
One round is one push. The gate is a chain, so the skill now says to re-run it
to completion locally before pushing rather than spend a round discovering the
next link (`fmt` failing means `clippy` never ran). The report gains a
`Classification:` line so a reader can tell a round that was spent from a round
that was refused.

## The pin (§3.5)

`kit_skills.rs` asserts both trees' `shepherd` names the four classes and the
four CRITICAL rows.

**Checked against pre-082 code:** the pin fails on the first needle
(`kit/shepherd: must mention "CRITICAL"`), so it is not asserting phrasing that
was already present. Verified by restoring `HEAD`'s shepherd, running the test,
and restoring.

## Why `extends`, not `amends`

048 §3.1 requires `shepherd` to "bound remediation to two rounds". This does not
change that bound; it narrows what may spend one. Additive, so `extends` on
`kit/.claude/skills/` (029), `.claude/skills/` (048), `kit_skills.rs` (048) and
`kit_embedded.rs` (065). No approved spec's text is edited.

## One deviation worth flagging

The parity line in the Verification block is `diff -r -x .DS_Store ...` rather
than the plain `diff -r` spec 081 uses. The plain form passes in CI, where a
fresh checkout has no such file, and fails on any macOS checkout. The excluded
name is gitignored and can never be part of what the kit ships, so nothing is
weakened. 081's line has the same latent false negative; not fixed here, since
that is 081's text.

## Gate

`lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (5 paths,
no drift), `cargo test --workspace --locked` (30 suites, 0 failures),
`clippy -D warnings`, `fmt --check`. `spec-spine verify 082` passes all ten
commands. Binary 0.18.0.

## Merge order

This branch and #170 both restale the whole index, so whichever merges second
conflicts on the 86 shards. The resolution is mechanical and the repo documents
it: merge the base, re-run `spec-spine compile && spec-spine index`, commit the
regenerated shards. `kit_embedded.rs` is generated by
`scripts/gen-kit-embedded.py`; it is never hand-edited.


bartekus:
Both review items addressed in 26e322c, though the first one turned out to be larger than the review framed it.

### 1. `.agents/skills/` — considered, and it is a finding against 081

The review guessed "separate governance domain, probably intentional." It is not. `.agents/skills/` is a **fifteen-skill copy of the pre-081 set with `.claude/` rewritten to `.Codex/`**, and it entered the repository in **spec 081's own commit** (`f5efb23`), whose §2 reads "No new file" while adding 15 tracked files.

```
$ git log --all --diff-filter=A -- .agents/skills/shepherd/SKILL.md
f5efb23 feat(081): the kit ships what the loop calls (#168)

$ diff <(git show f5efb23^:.claude/skills/shepherd/SKILL.md) .agents/skills/shepherd/SKILL.md
14,15c14,15
< `.claude/rules/orchestrator-rules.md` (checkpoints are real stops) and
---
> `.Codex/rules/orchestrator-rules.md` (checkpoints are real stops) and
```

It is unowned (`index owner` reports no owner), no `[index] extra_hashed_inputs` glob covers it, nothing in `AGENTS.md`, `CLAUDE.md`, `specs/`, `docs/`, `kit/` or the workflows refers to it, and the `.Codex/rules/` path its skills cite **does not exist** (`.codex/` holds `agents/` and `hooks.json`, no `rules/`). It also still carries the five skills 081 exists to remove. It passed the gate silently because an unowned `.md` trips neither C-001 nor C-002.

So 082 does **not** mirror into it. Doing that would make this spec the first to treat that tree as governed territory, which is a decision about whether it should exist at all. §4 now names it and D-3 records the reasoning, so the omission is visibly deliberate, which is what the review legitimately wanted. Reconciling or deleting it is its own spec, filed against 081.

### 2. The `adversarial-prompt-refusal.md` citation is correct

The review suspected the rule name implies it only covers refusing adversarial prompts. It covers both: the rule ends with "Two edits are always legitimate for the spec you are implementing: adding a file you created to its `establishes` list ... and recording a dated decision entry for a choice the spec was silent on." Spec **047** is titled *harness-rules-name-the-legitimate-edits* for exactly this.

The sub-point was fair, though: §3.3 described the HIGH coupling row without the citation the skill carries. It now cites it, so spec and skill agree.

### On Step 3b

Correctly identified as pre-existing and correctly not filed as a regression here. A green PR skipping the review-thread pass is a real gap in `/shepherd`'s routing, but it is 048 territory and not something 082 should absorb.

Gate re-run after the edits: `check`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple`, `kit_skills` (18 passed), `spec-spine verify 082` (10/10). All green.

bartekus:
Second review's `.agents/` conclusion accepted, thanks. Its one remaining item is a false positive, and it invents vocabulary the tool does not have.

**`status: draft` + `implementation: complete` is the correct and required state for this PR.**

The two are independent axes by design, and the suggested values do not exist:

```rust
// crates/spec-spine-types/src/frontmatter.rs
pub enum Status { Draft, Approved, Superseded, Retired }
```

There is no `accepted` and no `complete` status. `deny_unknown_fields` on the frontmatter grammar means either suggestion would fail the compile with exit 3, not merely mislead tooling.

`AGENTS.md` "Working the backlog" step 6 prescribes exactly this pairing: `/ship` opens the PR with `implementation: complete` once the `## Verification` block holds, and **"After merge, a second PR flips `status: draft` to `approved` (the ratify PR)."** A spec cannot be `approved` in the PR that builds it, because approval is a human act performed on merged work.

The claim that "most specs in the corpus set these consistently" reads the corpus at rest, where all 82 are approved and complete. At the equivalent moment in their own build PRs, the two most recent specs looked exactly like this one:

```
$ git show f514aeb:specs/080-.../spec.md | grep -E '^(status|implementation):'
status: draft
implementation: complete

$ git show f5efb23:specs/081-.../spec.md | grep -E '^(status|implementation):'
status: draft
implementation: complete
```

No change made. The ratify PR follows this merge, as the PR description says.

The Step 3b routing note is correctly classified as pre-existing and out of scope here; it is 048 territory.

bartekus:
**Pass 4 caught a real defect, and it was mine. Fixed in 605d058.**

The finding is exactly right: this branch was deleting all fifteen tracked files under `.agents/skills/`. Nothing intended that.

```
$ for c in $(git rev-list origin/main..HEAD | tac); do ... done
8e163d4 docs(082): file a refusal...              -> deletes 0
117421e chore(082): pending -> in-progress        -> deletes 0
d9d472d feat(082): shepherd classifies...         -> deletes 0
26e322c docs(082): name the .agents tree...       -> deletes 0
914c99b docs(082): make the rule citation...      -> deletes 15
```

The directory left the working tree between `26e322c` and `914c99b` (nothing in `scripts/gen-kit-embedded.py` deletes; it has no removal call), and a blind `git add -A` staged the removal without the staging list being read. Restored verbatim from `origin/main`; the branch now deletes nothing:

```
$ git diff --diff-filter=D --name-only origin/main...HEAD | wc -l
0
$ git diff --stat origin/main...HEAD -- . ':(exclude).derived'
 5 files changed, 426 insertions(+), 30 deletions(-)
```

The review's resolution path (b) was the correct one, and its reasoning for why is the same as the spec's: §4 says this spec "does not edit it, mirror into it, or assert parity with it" and calls it "not a territory 082 may quietly adopt." Deleting the tree is the loudest possible form of quietly adopting it, and whether it should exist is a human's call on its own spec.

## Why no gate caught this

Worth stating plainly, because it is the more important half. `couple` reported:

```
spec-spine couple: OK: 5 path(s) checked, no drift.
```

across the deletion of fifteen tracked files. Not a bug in `couple`: `.agents/` is unowned, so C-001 has no owning spec to demand an edit of, and `.md` is not in `coverage.rs::SOURCE_EXTS`, so C-002's `require_ownership` ratchet does not reach it either. From the gate's point of view those files were never governed, so removing them is not drift.

That is the same blind spot that let the tree enter the repository unowned in 081's commit. It now demonstrably lets it leave the same way. A tracked file that no spec claims can be added or deleted by any PR without the ownership ratchet noticing, and markdown is where that bites, because `require_ownership` is scoped to source extensions.

Filed as owed work, not fixed here: closing it means either widening `SOURCE_EXTS`, or a C-002 variant for tracked-but-unclaimed files of any extension, and either is a spec of its own with real blast radius across adopters.

Four passes in, this is the finding that justified the exercise. The gate verified the change; it could not verify what the change also did.

bartekus:
**Pass 5's second finding was right and in scope. Fixed in a684780.**

The CRITICAL stop ends the run before Step 3b, so the threads are never fetched, but the report template offered only `none | n addressed | n need a human`. A human reading `Review threads: none` on an escalation could not tell "there are no threads" from "nobody looked."

This is the same defect spec 080 fixed in the PR gate three days ago, in a different place: **a read that was not performed is not a clean verdict, and reporting it as one is a claim the run never established.** 080 taught the gate to say "I could not ask" instead of "stale"; this teaches the report to say "not read" instead of "none". Squarely in scope, because the CRITICAL stop is behavior 082 introduces.

- report line now reads `<none | n addressed | n need a human | not read: stopped at CRITICAL>`
- Step 2 states the thread line reads `not read` on that path, never `none`
- §3.4 requires the distinction, on exactly that reasoning
- the §3.5 pin asserts the phrase ships in both trees; the needle is new text, so it fails against pre-082 code
- Verification block gains the matching line, now 11 commands, all passing

On the first finding, the green-path routing skip: correctly read as acknowledged-and-deferred. Worth noting it is now *two* related gaps in the same skill, both about Step 3b being reachable only through failure, and the follow-up spec should take them together rather than one at a time.

On `status: draft` + `implementation: complete`: answered in full earlier in this thread with the enum and the two most recent specs at the same point in their own PRs. The lint does inspect both fields and passes; `spec-spine lint --fail-on-warn` is green, which is the direct answer to "worth resolving before merge if the linter inspects both fields together."

Gate after the change: `check`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple`, `cargo test --workspace` (0 failures), `clippy -D warnings`, `fmt --check`, `spec-spine verify 082` (11/11).

PR #174, opened 2026-09-10
docs(083): an attestation covers the territory it claims

bartekus:
## Summary

Files spec 083 as a draft. `spec-spine attest --spec <id>` cannot attest 14 of this corpus's 83 specs: `attest_spec` opens every resolved location with `read_to_string`, several unit spellings resolve to a directory rather than a file, and the read fails with `Is a directory`, so the verb exits 3 with no payload at all. The affected specs are 007, 008, 020, 029, 048, 051, 064, 068, 072, 074, 075, 077, 081 and 082: the distribution shims, the kit and the harness specs, which is where a claim is most often expressed as a subtree. 082 is on the list, which is how this surfaced, on its own ratify PR.

The dominant shape is not the obvious one. Thirteen of the fourteen fail on a plain `file` unit whose path is a subtree (`npm/`, `.claude/skills/`), because `resolve_unit` admits a file unit on a bare `exists()` check that is true of a directory. Only 048 fails on an explicit `directory` unit, and this corpus declares no owning `crate` unit at all. Spec 000 4.2 already defines the first two as one concept, so the draft keys its rule to the **resolved location** rather than to the declared kind: an implementation matching on `Unit::Directory` and `Unit::Crate` would leave thirteen of the fourteen still broken.

The specified fix walks a directory-resolved location instead of opening it, reusing the indexer's existing walker so the exclusion policy (`resolver_exclusions` plus `layout.state_dir`, which spec 039 3.5 says no exclusion entry may cancel) stays single-sourced, and folds the files into the same path-sorted `hash::content_hash` construction the file units already use. Two decisions are recorded rather than left implicit: the walk is deliberately not extension-filtered, because `SOURCE_EXTS` would hash zero files for a markdown-only subtree such as `.claude/agents/` and emit a confident hash over nothing; and a directory that walks to no files hashes its own path as a single piece, so an empty claim cannot attest as SHA-256 of the empty input, a constant every empty claim would otherwise share.

No `amends` edge. Spec 042 3.1 already requires a content hash for every owning unit without qualifying by kind, and spec 017 (2026-06-12) predates 042 (2026-09-06), so directory units existed when that sentence was written. This is conformance to the owning spec, the same shape as spec 070, not a change to what 042 requires.

Scope note: `attest --spec` also rejects the short id. That is deliberately **not** in this spec. `registry show 070` and `registry relationships 070` refuse it too, so the invariant specs 049 3.2 and 056 3.1 each assert in prose holds for two verbs of five and is enforced nowhere. Fixing it here would repair one verb and leave the false sentence standing in two approved specs. It is filed next as 084.

## Testing

The governance floor from `AGENTS.md`, all green on this branch:

- `compile` 0, `index` 0 (regeneration is byte-identical; the two shards are committed here)
- `check --fail-on-unresolved --fail-on-warn` 0
- `lint --fail-on-warn` 0 errors, 0 warnings, 0 info
- `index coverage --fail-on-untraced` 0 (80/80 specifically claimed)
- `couple --base origin/main --head HEAD` 0, no drift
- `cargo test --workspace --locked` all suites pass; `cargo fmt --all --check` clean. No `.rs` file changes in this PR.

The `## Verification` block parses into 13 commands under `verify --plan` and is not runnable yet by design: this is a draft filing, so `implementation` stays `pending` and the block's 3.1 to 3.4 assertions fail against current code, which is what makes them worth running after the build. Each was checked to fail for the right reason rather than vacuously: two comparison assertions require a non-empty reading first, because `test "$A" = "$B"` between two empty strings passes while proving nothing, and against current code both readings are empty. The scratch corpus was materialized and confirmed to compile, index, and exit 3 today, claiming one subtree in each spelling so the assertion also covers the kind-independence rule.

`status: draft` is deliberate. Approval is a human flip in a separate ratify PR.


bartekus:
Two of the three findings were actionable and are fixed in 4ea6c8b. The third is answered by the gate.

**Default `resolver_exclusions` (medium): valid, fixed.** `target` is in the compiled-in default (`crates/spec-spine-types/src/config.rs:191`), so the assertion passed today, but it was testing that default rather than the rule the spec states. The scratch corpus now writes its own `spec-spine.toml` naming `resolver_exclusions` before the assertion runs, so the 3.2 line tests the stated rule and still asserts something if the default ever moves.

**Cross-line filesystem state (low): valid, fixed.** The 3.2 line now removes its own `target/` directory before comparing, so the 3.4 baseline reads the same tree 3.3 did. The preamble also now states the constraint rather than leaving it inferred: spec 049 3.2 forbids a line depending on a variable another line set, these lines share a filesystem instead, and that is only safe while each mutation is undone by the line that made it.

One correction to the reasoning, which does not change the fix. The failure mode described as silent is loud: `verify` stops at the first failing command (spec 049 3.2), so a broken exclusion fails the 3.2 line and 3.4 never runs to pass for the wrong reason. The fix is still right, for the first reason rather than the second.

**`extends` edges with no paired implementation (informational): no change needed.** The gate answers it directly. `couple --base origin/main --head HEAD` exits 0 locally, and the `self-governance (compile · index · lint · couple)` job passed on ffdd554 with these edges present. The coupling gate fires on changed claimed paths, not on declared edges pointing at unmodified units, so a draft that claims territory it has not written is not a refusal. That is the same window specs 044 and 045 define for an in-flight spec; the unresolved-unit half would surface as `W-001` under `check --fail-on-unresolved`, which also passed.

bartekus:
Second pass: one finding was load-bearing and is fixed in e12ac14, one is refuted by measurement, one is a fair readability point and is fixed.

**Symlink behaviour (the real one): valid, fixed.** Confirmed against the code rather than taken on faith. `walk` recurses on `path.is_dir()` (`crates/spec-spine-core/src/index.rs:1191`), and `is_dir` follows links, so both risks named are real: a cycle inside a claimed subtree diverges, and a link out of the tree would hash content this repository does not own into a payload whose purpose is attesting what it does own.

3.2 now requires that a symlink is neither traversed nor dereferenced, and contributes exactly one piece: its own repo-relative path with the link target text as content. That was chosen over the two alternatives, and D-4 records why. Following links is the option that can hang. Skipping links entirely is safe but contradicts 3.4, since adding or retargeting a link inside claimed territory would leave the hash unmoved and the change invisible. Hashing the target text keeps both properties and matches how git stores a symlink, as a blob whose content is the target path, so the attestation records the same fact the repository does. 3.5 gains a guard for it, including a link whose target is a directory.

The existing walker's own behaviour is now scoped out in 4 rather than left ambiguous: `walk_source` serves the resolution and coverage walks, and whether those should change is a question about spec 004's territory that wants a spec which has measured it.

**`depends_on` incomplete relative to `extends`: not a defect here.** Measured across the corpus: of the 63 specs carrying `extends` edges, **43 name at least one target absent from `depends_on`**. That is the convention, not an omission. Spec 042 itself, the spec this one extends, lists `extends` targets 001 and 000 while declaring neither in `depends_on`. The two keys answer different questions: `depends_on` gates schedulability in `registry plan` (spec 038), and both 001 and 004 are approved with `implementation: complete`, so declaring them would change no verdict. `lint --fail-on-warn` reports nothing here, and no L- rule requires the pairing. If the pairing should be required, that is a lint rule and a spec of its own, applied to 43 existing specs rather than to this one.

**Sweep diagnostic: fair, fixed.** The all-83 loop no longer discards stderr; it names the spec that failed. Section 1's entire claim is which specs are unattestable, so a sweep that hides the identity was the wrong shape.

Portability note, agreeing with your own conclusion: `exit` inside a `for` body under `sh -c` is consistent across bash and dash, so the assertion was already portable. The change is about the diagnostic, not the exit code.

bartekus:
Third pass: two applied in ff77c0c, one declined with a measurement, one acknowledged.

**Symlink guard unverifiable from the CLI block: valid, and the most useful finding in this review.** You are right that `cargo test --test attest` cannot police a test that was never written, and an absent test is not a failing one. That is a vacuous guard, and this corpus has a spec about exactly that failure mode. The block now proves the rule itself rather than delegating it, twice:

- a symlink moves the hash, which fails if links are skipped;
- a symlink cycle still terminates at exit 0, which hangs or fails if the walk descends through it.

Both were run against current code before being written down: each exits 1, so neither passes vacuously, and the cycle line does not hang today because `attest` exits 3 before reaching any walk.

**Prose rule contradicted by 3.4: valid, fixed.** The rule was right and the block did not keep it. The 3.4 assertion now removes `d1/f.txt` before comparing, and I verified the scratch tree is empty after the two new symlink lines rather than assuming it.

**Shared scratch path: the concern is fair, the proposed fix does not work.** `/tmp/ss083-$$` would give each line a *different* directory. Spec 049 3.2 runs every fence body line as its own command, so `$$` is the pid of that line's shell, not a run-wide constant. Measured rather than reasoned about: two consecutive `sh -c 'echo $$'` invocations print 10551 and 10552. Under that form the corpus would be created by one line and read by another that cannot see it. A run-wide unique path needs a value that survives across lines, which is precisely what 049 3.2 forbids, so the fixed path is not a shortcut here but the only shape the rule permits. `verify` is also an on-demand verb rather than a CI job (spec 049 keeps it out of the gate chain deliberately), so two concurrent runs on one machine is not the environment it targets.

**Coupling note for the implementing PR: acknowledged, no edit.** Worth one correction for whoever picks it up: a subset does not require waivers. The implementing change advances 083 itself, and `couple` refuses a changed claimed path whose *owning spec* did not change, so the four `extends` edges are covered by that same spec edit whether the change touches all four units or two of them.

Unless the next pass finds something that changes what the spec requires, this is where I stop iterating and merge on green.

bartekus:
Fourth pass: both findings applied in a5f53da. This is the last iteration; I merge on green.

**Null-tolerant `contentHash` match: valid, and worse than described.** You are right that `sort -u | wc -l` returns 2 for one real hash plus one null. The second-order effect is the more serious one: the loose pattern also defeated the non-empty guard the other comparisons depend on, because `"contentHash": null` is a non-empty reading, so `test -n "$A"` would have accepted an all-null payload and compared nothing against nothing. All seven readings now require the opening quote of the hash string.

**Symlink assertion consistent with both behaviors: valid, and fixed with an assertion rather than a note.** You are right that descending into the link's target would move the hash too, so that line proved recording but not non-traversal. Rather than only documenting the gap, there is now a third assertion that discriminates: the link points outside both claimed subtrees, and editing that target's content must leave the hash unchanged. Under the required behavior the target text is what was hashed, so the edit is invisible; under traversal the hash moves. The preamble now states what each of the three proves and what none proves alone.

**Evidence, since the point of this block is that it fails first.** All eight assertions were run against current code exactly as `verify` would run them: every one exits 1, so none passes vacuously. The scratch tree afterwards contains exactly `d1`, `d2` and `spec-spine.toml`, which also confirms each assertion undoes its own mutation on the failing path, not just the passing one. The corpus-setup lines are the single exception and the preamble now names them instead of claiming a rule the block did not keep.

Thanks for four useful passes. Two of the findings across them were the same class of defect and worth naming, because it is the one this corpus keeps hitting: an assertion that cannot fail proves nothing, whether it is a unit test that was never written or a grep that matches the null it exists to exclude.

bartekus:
Fifth pass: the state-root finding was worth breaking my own "last iteration" line for, and it is fixed in 7aaf8b2. This is genuinely the last round; I merge on green regardless of what a sixth pass says, unless it changes what the spec requires.

**`layout.state_dir` never exercised: valid, and the sharpest finding of the five.** 3.2 calls that pruning not optional and argues no `resolver_exclusions` entry can express it, and then nothing in the block could fail if an implementation ignored it. That is a MUST with no way to fail, which is the same defect this PR has now corrected in three different shapes: a unit test that might never be written, a grep that matched the null it existed to exclude, and now a clause with no assertion at all. Merging it while having argued that point twice in this thread would have been incoherent.

The assertion needed the state root placed **inside** a claimed subtree, at `d1/state`, since that is the only placement where pruning is observable in a unit's hash, and where no `resolver_exclusions` entry could have produced the same effect (that list matches directory names, not path prefixes, which is 3.2's own argument for why the two mechanisms are distinct).

**`d2/` emptiness precondition: valid, stated.** You are right that the recording assertion is only discriminating while `d2/` is empty, and that nothing prevented a future edit from populating it. The prose now names it as a precondition of the block rather than an incidental detail, and says what breaks if it changes.

**Cleanup skipped on mid-block failure: correct, and now stated as a decision.** `verify` stops at the first failure, so the final `rm -rf` does not run. That is deliberate rather than overlooked: the tree is what a reader needs to diagnose the failure, and the next run's leading `rm -rf` means stale residue can never influence a later verdict. Written down instead of left to be rediscovered.

Nine assertions now, all run against current code before being committed: every one exits 1, and the scratch tree afterwards holds exactly the two claimed directories and the config.

bartekus:
Sixth pass: one applied in 250ad96, one refuted with evidence. Merging on green.

**3.3 path normalization: valid, and the one kind of finding I said would earn another round.** This is a requirement-level ambiguity rather than a verification one. That piece is the only one in the construction whose path is not carried in from a walked file, so it is exactly where an implementation could reach for the absolute path the resolver holds, and an empty directory would then hash differently under `/home/alice/proj` than under `/home/bob/proj`. A machine-dependent hash defeats 3.4 and spec 042 3.5. The clause now says repo-relative POSIX explicitly, with the reason recorded beside it so the next reader does not have to reconstruct it.

**`establishes:` frontmatter key: this is the correct key, and the tests already proved it.** No change. `establishes` is the ownership key in `standards/spec/templates/spec-template.md`, in `standards/spec/contract.md` under "Typed edges", and in spec 000 4.2; there is no `owns:` or `units:` in this grammar. More directly, the scratch corpus was run before the assertions were written down, and the pre-083 failure it produces is:

```
io error: read .../ss083/d2/ for spec '001-dirs' unit Directory { path: "d2/", planned: false }: Is a directory
```

That message can only be produced by a resolved, owned unit, so the key parsed and the units reached the resolver. If `establishes` were passed through as extra frontmatter under spec 013, the payload would carry `units: []` and the error would not name a unit at all.

Flagging this one specifically because it is a repeat: an earlier run of this job reported `establishes` as an unknown field and suggested an `owns:` key that does not exist in this grammar. The structural findings in these reviews have been consistently good, and five of them improved this spec, but existence claims about the frontmatter vocabulary are worth checking against `frontmatter.rs` and the template before acting on them.

PR #177, opened 2026-09-11
docs(084): a short id names the same spec at every verb

bartekus:
Files spec 084 as a draft (`status: draft`, `implementation: pending`). Nothing
is implemented here; the diff is the spec and its two regenerated shards.

```
$ git diff --stat origin/main...HEAD -- . ':(exclude).derived'
 specs/084-a-short-id-names-the-same-spec-at-every-verb/spec.md | 567 +++
```

083 4 named this as the next spec to file.

## The defect

Six arguments take a spec id, and they disagree about what one is. Measured on
0.18.0 at `6fca5ee`:

| argument | `070` today |
|---|---|
| `compile --spec` | accepted (056) |
| `verify` | accepted (049) |
| `registry show` | exit 1 |
| `registry relationships` | exit 1 |
| `attest --spec` | exit 1 |
| `verify-attestation --spec` | **exit 3**, reading `by-spec/070.json`, with a hint to run a command that refuses the same argument |

Two approved specs describe this as already solved: 049 3.2 says `verify 049`
"resolves as `registry show 049` does", and 056 3.1 says "as everywhere else the
short form is accepted". Nothing enforces either sentence.

Found while filing, and written into the spec:

- **Four copies of the policy, not three.** 083 3.2 missed
  `compile.rs::resolve_spec_ref`.
- **The strict copies join the argument onto a path** for the exact case, so
  `verify ../specs/<id>` resolves at exit 0 and `compile --spec ./<id>` reports a
  false `V-001` on a valid spec.
- **Three traps for a partial fix.** Resolve in `show` alone and `relationships`
  computes incoming edges against the raw argument, printing an empty
  `depended_on_by` at exit 0. Resolve only `attest_spec`'s record lookup and its
  index-mapping lookup finds nothing, so the payload lists zero units with
  `resolution.ok: true`. Resolve all of core and the CLI still names the file
  `by-spec/070.json`.

## What 084 specifies

- One public policy in a new `spec_id.rs` that all six arguments and both
  frontmatter-reference resolvers call (the maintainer chose to fold the
  reference family in).
- Each verb resolves against the ids it already reads: disk for `compile --spec`
  and `verify`, the committed registry for `show`/`relationships`, the in-memory
  compile for `attest`, the attestation files for `verify-attestation`.
- Resolve once, then only the resolved id: comparisons, file names, emitted
  fields.
- Enforcement, not prose: a six-argument matrix test, and a clap census unit
  test that fails when a seventh id argument appears outside it.

## Edges

- **`amends` 016**, for its section 2, which records the compile resolver as "a
  local mirror ... rather than a shared call". 016's reason for the mirror (the
  compile gate must not depend on the indexer's file) is kept: the shared module
  belongs to neither 001 nor 004. 016's text is not edited (spec 040).
- `extends` on ten units owned by 001, 002, 004, 023 and 049. No `amends` on
  002, 042, 049 or 056 (D-2): each states behavior 084 makes true rather than
  behavior it changes.
- The three new files are claimed `planned: true` (spec 076) until the build
  writes them.

`risk: high`: the lenient adapters sit on the path that writes committed
registry and index shards. 3.6 requires those to stay byte-identical, and
`spec-spine check` is the witness.

## The Verification block was run against current code

Every line was run one at a time on this branch, before any implementation:

- all 15 lines labelled "fail first" exit non-zero (the three test targets, the
  four short-form acceptances, the four byte-identity and file-name lines, the
  path-is-not-an-id line, the single-copy line, and the one-message lines for
  ambiguity across six arguments and no match across five);
- all 6 regression guards and the 3 setup lines exit 0.

Three vacuous passes are closed on purpose: a `cargo test` filter that matches
nothing, `test "$A" = "$B"` between two empty strings, and a regression that
breaks the short and full forms alike.

## Review before filing

An independent read-only review checked every code claim and every cross-spec
citation against the tree and found none false; it reproduced the section 1
transcripts live. It found four gaps, all fixed in this commit:

- the full-id guard covered five arguments, not six (`verify-attestation` added);
- no line checked that the step-4 refusal is one message (the no-match line added);
- `verify_attestation.rs::validate_spec_id`, the existing guard on a path-shaped
  argument, was unmentioned, and a rewrite could have validated only the resolved
  id and reopened it on the exit-3 fall-through (3.2 now requires it to run on the
  raw argument first, with a guard line);
- the "no shared shell state" rule was cited to 049 3.2 alone; it is 3.2 and 3.5.

It also confirmed `init`'s ordinal-collision warning is not a copy of this policy,
which section 4 now says.

## Gate

```
$ spec-spine check --fail-on-unresolved --fail-on-warn   # exit 0, both trees fresh
$ spec-spine lint --fail-on-warn                          # 0 error, 0 warning, 0 info
$ spec-spine index coverage --fail-on-untraced            # 80/80, exit 0
$ spec-spine couple --base origin/main --head HEAD        # 1 path checked, no drift
$ cargo test --workspace --locked                         # 30 suites, 604 passed, 0 failed
$ cargo clippy --workspace --all-targets --locked -- -D warnings   # clean
$ cargo fmt --all --check                                 # clean
```

`spec-spine registry plan` now lists 084 as ready, with its three planned units.

## Approval

`status: draft` is deliberate. Merging this PR files the spec; it does not
approve it. Per `AGENTS.md` "Working the backlog", this repository builds a filed
spec while it is still `draft` once a human names it (`/build 084`), and a
separate ratify PR flips it to `approved` after the build merges.


bartekus:
Dispositions for the ai-review, each checked against the tree. Fixed in ee321d0.

1. **No-match uniformity, `compile --spec`: premise false, gap real, fixed.**
   Every one of the five verbs writes its refusal to stderr as one line today
   (`compile --spec 999` prints `spec-spine: not found: spec '999'` to stderr
   and nothing to stdout). But the underlying vacuous path is real: a verb that
   wrote no refusal at all would add nothing to the file and leave the distinct
   count at one. Both one-message lines now require exactly one stderr line per
   argument (5 and 6) before counting distinct lines.
2. **Scratch frontmatter missing `kind`/`owner`/`implementation`/`risk`: not a
   defect.** The contract's required set is `id`, `title`, `status`, `created`,
   `summary` (`standards/spec/contract.md`, "Required frontmatter"). The setup
   line already asserts the outcome it depends on, exit 1 from `V-004` **and**
   the presence of `by-spec/001-a.json`, and it passes; a schema change would
   fail that line rather than the later ones.
3. **`verify-attestation` in a corpus with no `spec-spine.toml`: not a defect.**
   All config keys are optional and an absent file means defaults. Measured:
   `verify-attestation --spec 001-a --recompute` against that scratch corpus
   reaches `recompute: MATCH`.
4. **`validate_spec_id` wording: fixed.** 3.2 now states it as written:
   presence tests for `/`, `\` and NUL, equality tests for `.` and `..`, the
   latter sufficient because without a separator `..` can only name a parent as
   the whole argument.

The Verification block was re-run line by line against current code after the
change: the same 15 lines fail first, the 6 guards and 3 setup lines pass.

bartekus:
Dispositions for the second ai-review, checked against the tree. 0e0d3f7.

1. **`compile` writing shards on a V-004 corpus: specified, and not silent.**
   001 3.5 has `compile` write the registry and then exit by the verdict, and
   the registry records it as `validation.passed: false` (001 3.2). The setup
   line asserts both halves (exit 1 and `by-spec/001-a.json` on disk), so a
   compile that stopped writing on failure fails that line. It could not produce
   a false pass later either: `registry show` over a missing registry exits 3
   (measured), which fails the ambiguity line's exit-1 assertion. Writing the
   shards by hand would be the hand-edit constitution II forbids. The
   Verification prose now says all of this.
2. **`wc -l` on an unterminated line: not reachable today, changed anyway.**
   Every refusal goes through one `eprintln!` in `main.rs`, which terminates the
   line, and an undercount would be a loud failure rather than a vacuous pass.
   Both counts now use `grep -c .`, which costs nothing.
3. Agreed, no bug (as the review says).
4. **`attest --json`: exists.** `attest` has taken `--json` since spec 037
   (`main.rs`, `Command::Attest { json }`; listed among the verdict verbs in the
   README). `attest --spec <full id> --json` prints the envelope today; the
   comparison line's `contentHash` guard is what keeps an empty reading from
   passing.
5. **Amending 016 without editing it: that is spec 040's rule.** An `amends`
   edge is declared once, in the amending spec, and the amended `spec.md` is
   not edited (`standards/spec/contract.md`, "Amendment authoring"). The inbound
   view is compiled: on this branch `registry relationships 016-short-id-resolution`
   already reports `amended_by (incoming): 084-...`. D-1 now cites 040.

Re-run line by line against current code after the change: the same 15 lines
fail first, the 6 guards and 3 setup lines pass. I'm stopping the review loop
here; further ai-review passes are yours to weigh.

bartekus:
Dispositions for the fourth ai-review. 05a8c43.

1. **3.3's attestation-file bullet stated absolutely: fixed.** It now reads: the
   file is named by the resolved id when the argument resolves; when it matches
   no file there is no resolved id, and the name comes from the argument as
   given (step 4's fall-through, 3.2 and D-4), after `validate_spec_id` has
   passed it.
2. **Step 4 at `verify-attestation` unasserted: already asserted; the line now
   says so.** At this verb the set is the attestation file stems (3.2), not the
   corpus. The guard deletes `by-spec/016-short-id-resolution.json` first, so
   `016` matches no stem and takes the fall-through: an implementation that
   exited 1 for every non-resolving id would fail that line (it asserts exit 3).
   The premise that "016 always resolves" holds for the corpus-backed verbs, not
   this one. The line now carries its own comment naming the path it exercises,
   rather than adding a second line that proves the same thing.

The previous head (40c1e0d) was green on all nine checks. Re-run line by line
against current code after this change: the same 15 lines fail first, and the
6 guards and 3 setup lines pass.

bartekus:
Fifth ai-review: no bugs, nothing blocking. Three LOW prose notes, acknowledged and deliberately not acted on before merge, since each push draws a fresh review:

1. The step-4 guard leaves `by-spec/016-short-id-resolution.json` removed. It is an on-demand, gitignored attestation that `attest --spec 016-short-id-resolution` regenerates, and the post-merge `/verify` runs in a clean checkout where it does not exist anyway. The ordering is sound, as the review notes.
2. The no-match line already counts its stderr lines (`L -eq 5`); the prose names the count for both one-message lines ("The two one-message lines also count their stderr lines").
3. A pre-084 `by-spec/016.json` is unreachable, as the review says: pre-084 `attest --spec 016` exits 1 before writing.

Head 05a8c43 is green on all nine checks. Merging.

PR #180, opened 2026-09-12
chore(089): ratify (draft -> approved)

bartekus:
089 was built and merged as `2ecd8c1` (#178) and has sat at `status: draft`,
`implementation: complete` ever since. That pair is the lifecycle discrepancy
the revision-3 packet asked to resolve "without assuming a merge approved the
spec". It did not. This is the approval.

## Evidence gathered before the flip

Measured at `0e41641` on `spec-spine 0.18.0`:

| case | old shape | shipped shape |
|---|---|---|
| manifest present, command fails | prints `no Cargo.toml, skipping`, **exit 0** | no output, **exit 1** |
| manifest absent | skips, exit 0 | skips, exit 0 |

The second row matters as much as the first: spec 064's specify-first no-op
still holds, so the fix bought the failure signal without costing the skip.

- `kit/Makefile` carries no `test -f M && cmd || echo` shape, and all five
  guarded recipe lines are an explicit `if`.
- `spec-spine verify 089` passes all five declared commands.
- `cargo test -p spec-spine-core --test kit_gate` passes 10 of 10, including
  `a_guarded_recipe_skips_when_absent_and_fails_when_the_command_fails`, which
  runs each shipped line with its command forced to `false`.

## Gate

`check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`,
`index coverage --fail-on-untraced` and `couple` all exit 0; `cargo test`,
`clippy -D warnings` and `fmt --check` are green. Only 089's two shards move,
since a `spec.md` edit restales its own shards and nothing else.

## Adopter note for the release

This fix is **not** delivered by bumping a version pin. The defect is in the
file adopters copy, so the release notes have to tell them to re-copy
`kit/Makefile`.

No waiver is needed or claimed.

### DIFF
--- a/.derived/codebase-index/by-spec/089-a-skip-and-a-failure-are-different-answers.json
+++ b/.derived/codebase-index/by-spec/089-a-skip-and-a-failure-are-different-answers.json
@@ -60,8 +60,8 @@
       }
     ],
     "specId": "089-a-skip-and-a-failure-are-different-answers",
-    "specStatus": "draft"
+    "specStatus": "approved"
   },
   "schemaVersion": "1.1.0",
-  "shardHash": "1a8f107be3e241300aeafc76899889029c9fbb4542490a501508e5fd09a34f65"
+  "shardHash": "c249230bfb7712e2b2f5f61d8c2b04bf4dfe92f617ee8a04202080bcd92b3f0e"
 }
--- a/.derived/spec-registry/by-spec/089-a-skip-and-a-failure-are-different-answers.json
+++ b/.derived/spec-registry/by-spec/089-a-skip-and-a-failure-are-different-answers.json
@@ -49,10 +49,10 @@
       "Verification"
     ],
     "specPath": "specs/089-a-skip-and-a-failure-are-different-answers/spec.md",
-    "status": "draft",
+    "status": "approved",
     "summary": "The kit's Makefile guards its language targets on a manifest probe, written as `test -f M && cmd || echo skipping`. That is not an if/else: the `||` branch fires when the probe is false OR when the command fails, so on a repository that HAS the manifest a failing command exits 0 having printed \"no manifest, skipping\". Any adopter who copies `kit/govern.yml` unmodified into a repository with a `Cargo.toml` gets a build job that cannot fail: a broken `cargo test`, a `cargo fmt --check` diff and a `clippy -D warnings` hit are all reported as a skip and the check goes green. Spec 064's acceptance could not catch it, because it asserts the probe is a manifest probe and that is true of the broken shape and the fixed one alike. This spec makes the guard an explicit `if`, and adds the acceptance that tells the two apart: the shipped recipe lines are run in both states.\n",
     "title": "A skip and a failure are different answers"
   },
-  "shardHash": "b3708ac32cd646a2d52c7e71affda5a6e8eda559ecb6c408d31efaa27338ef8e",
+  "shardHash": "e2a8de53745e48d8928974a5b08e213786bc2cd0450e94f708de1341576d9f1b",
   "specVersion": "1.2.0"
 }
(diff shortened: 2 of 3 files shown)


bartekus:
Both findings are false positives. Evidence for each, since the review asked for confirmation rather than asserting a defect.

### Finding 1: an attestation cannot accompany this, by design

No attestation can appear in any diff in this repository. `.derived/attestation/` is gitignored (`.gitignore:30`), and spec 042 makes that a deliberate choice rather than an omission. From 042's own summary:

> It stays on-demand and gitignored like 023's, **deliberately**: the design note wanted a committed per-spec bundle, and committing one would restale on every edit to any claimed unit and would need its own freshness gate, which buys churn rather than assurance.

So 042 requires no attestation artifact at approval time, and the suggested remedy ("or add one") would contradict the spec it cites. `git ls-files | grep attestation` returns only source and spec files; no attestation record is tracked anywhere.

The precedent is identical in shape: `chore(083): ratify` (`e4032ff`, #176) is the same three files, two shards and one `status:` line.

### Finding 2: 041 is satisfied mechanically, not by assertion

Spec 041 is the spec that **describes this PR shape as the corpus norm** and then closes the window inside it. From its summary:

> A spec is filed as `draft` with `implementation: complete` when its code lands (spec 036 did exactly that at `46474a3`) and stays that way until a later ratification PR flips it to `approved`, so every spec passes through a period where its strongest claim about code is the one claim nothing checks.

041's mechanism is that `implementation: complete` makes an unresolved owning unit a **refusal** instead of a counted `W-001`. That flag was set by #178, not by this PR, and the refusal is `check --fail-on-unresolved`, which exits 0 on this branch and is run by the `self-governance` job that is green above. The claim is not merely verifiable; it is enforced on every run.

On "the PR description provides no evidence": the description carries the measured old-shape versus shipped-shape table, records that `spec-spine verify 089` passes all five declared commands, and names `kit_gate` at 10 of 10, including `a_guarded_recipe_skips_when_absent_and_fails_when_the_command_fails`, which runs each shipped recipe line with its command forced to `false`.

### One real issue, self-inflicted and fixed

The first push carried an **unsigned commit**, which is why `mergeStateStatus` read `BLOCKED` despite all nine checks passing: `main` has `required_signatures: true`. The commit has been re-signed and the branch updated. Nothing about the content changed.

No waiver is needed or claimed, and no spec other than 089 is touched.

bartekus:
Second review's ask satisfied. Both verbs run against this sha, with output.

**`spec-spine verify 089`**: `passed (5 command(s))`, exit 0. The five are 089's own declared acceptance, including the two that tell a skip from a failure by running each shipped recipe line with its command forced to `false`.

**`spec-spine attest --spec 089-...`**: exit 0, `attestationHash 3568cd2c...`. The payload witnesses all three units 089 claims, each with a content hash:

| unit | |
|---|---|
| `crates/spec-spine-core/src/kit_embedded.rs` | `367bad2c...` |
| `crates/spec-spine-core/tests/kit_gate.rs` | `5363bd71...` |
| `kit/Makefile` | `53a1f300...` |

`verdicts.compile.ok`, `verdicts.lint.ok` and `verdicts.resolution.ok` are all true, and `lifecycle` reads `{status: approved, implementation: complete}`, which is the pair this PR creates. `verify-attestation --spec ... --recompute` returns `MATCH` at exit 0.

The attestation is not in the diff because it cannot be: `.derived/attestation/` is gitignored by spec 042's deliberate design, and the working tree is unchanged by having written it. That is the same point as the reply above, now with the artifact itself as evidence rather than the argument.

One correction for the record, offered rather than argued: 042 governs `attest --spec`'s payload and does not require an attestation at approval time. 041's requirement is that `implementation: complete` turns an unresolved owning unit into a refusal, which `check --fail-on-unresolved` enforces on every CI run. Neither spec gates a `status:` flip on a witnessing step. The witnessing is worth doing anyway, which is why it is above.

PR #185, opened 2026-09-13
docs: the docs describe 085 and 086 as shipped; build identity is documented

bartekus:
## Summary

The pre-release docs sweep for v0.19.0. Nothing here changes behavior; it
brings the prose in line with what specs 085 and 086 shipped, and records the
maintainer's decision on design note 04 D1.

- **`website/docs/cli/index.md`** described `index check` as a content-hash
  comparison and showed an invented `Expected hash abc123...` output. Since spec
  086 it compares committed shard bytes with a fresh in-memory index. The page
  now names the three drift classes (`modified`, `missing`, `orphaned`) and the
  `blocking-diagnostics` line, and the example is output captured from a real
  run of a 0.19.0 build against a tampered copy of this repository's index.
- **`coupling-gate.md`, `overview.md`, `docs/concept.md`** repeated "detecting
  staleness by content hash"; each now says byte-for-byte comparison.
- **`docs/authority-evidence.md`** told consumers to hash attestation bytes
  themselves "until draft spec 085 lands" and called 086 a draft. The
  measurements at `75181a5` are kept as taken; §4, §5 and §8 now say what
  v0.19.0 does beside them. §8's new column is spec 085's own `## Verification`
  block, which passed in full (25 commands) on a 0.19.0 build, and the two
  refusal messages quoted were checked by hand.
- **Build identity (design note 04 D1)**, decided for v0.19.0: document the
  limit rather than stamp a source commit or give `main` a pre-release version.
  The consumer doc's §2 states it (a version string names a release only for a
  tag or distribution-channel build; a consumer accepting other builds records
  the commit itself), and D1 and the F3 row record the decision.

Every changed path is unowned (`spec-spine index owner` reports no owner for
each) and on the bypass floor, so no spec is edited and no waiver is needed.

## Testing

`spec-spine check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`,
`index coverage --fail-on-untraced` and `couple --base origin/main --head HEAD`
all exit 0; `compile` and `index` leave the committed trees unchanged.


bartekus:
Answering ai-review's third pass (no change made; each checked against the source):

- **Coupling on `website/docs/cli/index.md`**: `spec-spine index owner website/docs/cli/index.md` reports no owner, so spec 086 does not claim it, and the self-governance job's `couple` passed on 539454d. Same answer as the two earlier passes.
- **`slices.json` sidecar**: correct. `check_slice_freshness` reads `read_committed_slice_hashes` (`crates/spec-spine-core/src/index.rs:827`), which reads `slices_path`, documented as "the per-slice sidecar (spec 012, relocated by spec 024): `slices.json`".
- **§4 pointer**: §4 is "What each digest covers" (line 205), and the edited "An index shard's hash does not cover its own body" bullet is in it (line 265). §5 and §8 were edited as well.

PR #187, opened 2026-09-13
feat(088): delta classifies a change under the merge base's rules

bartekus:
## Summary

Builds spec 088. `spec-spine delta --base B --head H [--json]` resolves `merge-base(B, H)`, exports the merge-base and head trees into two temporary directories (through a private index file, removed afterwards), and classifies every changed path into `implementation`, `requirement`, `verification`, `authority`, `lifecycle`, `constitutional`, `policy`, `derived`, `bypassed`, `unowned` or `unknown`, under the **merge base's** configuration and committed index. The report carries both plan digests for a spec whose acceptance changed, both owner sets for a path whose ownership moved, lifecycle key changes, per-class counts, and `priorPolicy`. It is a record, not a gate: exit 0 whenever a report was produced.

- Both §1 candidates that pass `couple` at exit 0 are now reported: weakening one's own `## Verification` block is `verification` (base and head plan hashes), and a new `extends` claim is `authority` on the unit (`baseOwners [001-a]`, `headOwners [001-a, 002-b]`).
- A candidate's `spec-spine.toml` edit is `policy` and does not reclassify the rest of its diff (D-1).
- Facade `delta_json`; `DELTA_SCHEMA_VERSION` 0.1.0; the envelope gains the `delta` verb token, so `VERDICT_SCHEMA_VERSION` moves to 0.4.0 (additive MINOR, spec 037).
- `verify.rs` gains `without_verification_section`, so "body outside `## Verification`" uses 049's one heading grammar (`extends` 049).
- Docs: `docs/authority-evidence.md` (the new question row, the version axes, what `required: false` does not mean, §7), `docs/api.md`, `README.md`.

### Decisions recorded for ratification (the spec was silent)

- **D-5** head owners come from the head tree indexed under the base's configuration; the head's committed index is never read.
- **D-6** a stale base index is exit 2 with no report.
- **D-7** the merge base's own configuration must load (exit 3); a head configuration that fails to load is `policy` + `unknown`.
- **D-8** the "kinds" of §3.3 are path kinds; `unowned`, like `implementation`, excludes them. **Worth a look at ratification:** read without that exclusion, every `spec.md` is `unowned`, and a spec change no spec row places (a frontmatter comment, prose inside `## Verification`) would read `unowned` with `required: false`, which is the outcome D-4 is written against. Such a change is `unknown` instead.
- **D-9** changed paths by `git diff --name-only -z --no-renames` (couple's unified-diff parser drops binary and mode-only changes); trees via `read-tree` + `checkout-index`, not `git archive` (honors `export-ignore`).
- **D-10** per-change `change` (added/deleted/modified), token-ordered class lists, every class counted.
- **D-11** `policy` is what `global_inputs_hash` folds, found with its own glob walk (shared from `shard.rs`, `extends` 024) over both trees. The first build used a string pattern test; an independent review reproduced it labeling a file under a bare `kit/**` `policy` while editing it left `check` fresh (spec 069's dead glob). Fixed in `892a35b` with a test that holds each verdict to `check_index_freshness`.

## Testing

- Gate as `AGENTS.md` lists it, on `HEAD`: `compile`, `index`, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced` (86/86), `couple --base origin/main --head HEAD`, `cargo test --workspace --locked`, `cargo clippy ... -D warnings`, `cargo fmt --all --check`: all exit 0. `cargo test -p spec-spine-core --no-default-features --locked`: 457 passed.
- `spec-spine verify 088-a-change-is-classified-under-the-bases-rules`: passed, 15/15 commands, clean tree.
- `crates/spec-spine-core/tests/delta.rs`: the §3.8 matrix plus the D-4, D-8 and D-11 cases (17). `crates/spec-spine-cli/tests/cli.rs`: git end to end (base rules with the head checked out, a binary file, commit ids echoed, repo and temp dirs left clean, exit 3 envelope, stale base exit 2).
- Fail-first probes, each reverted: reading the head's committed index fails 3 matrix cases; reading the working tree's configuration fails the CLI base-rules case; the facade reading the head's configuration fails the facade case; the pre-D-11 string matcher fails the D-11 case.
- Independent review of the diff: one confirmed finding (D-11, fixed). Unconfirmed and not addressed: `checkout-index --prefix` with a Windows path (CI runs tests on Linux only), and a non-UTF-8 file name decoded lossily from `git diff --name-only -z`, which `couple`'s parser also does.
- Dogfood on this branch: `delta --base origin/main --head HEAD` reports 19 paths in 0.5 s, `priorPolicy` required for `authority, lifecycle, requirement`.

Found, not fixed here (outside 088's territory):

- The `## Verification` policy line does not discriminate D-1: `src/a.rs` is explicitly claimed, and a claim outranks a bypass prefix (spec 009), so it classifies `implementation` under either configuration. The CLI test adds an unclaimed path that does.
- `couple` registers paths from `+++`/`---` headers, which git omits for binary and mode-only changes, so neither reaches C-001 or C-002.


bartekus:
Reply to the AI review, point by point, checked against the source:

1. **Coupling on `cmd_couple.rs`: not a violation.** 088 declares `{ spec: "005-coupling-gate", unit: "crates/spec-spine-cli/src/cmd_couple.rs", nature: additive }` in its `extends` list (spec.md line 39), and `spec-spine index owner crates/spec-spine-cli/src/cmd_couple.rs` lists `088 ... (extends)` beside `005 ... (establishes)`. The same holds for `main.rs`, both `lib.rs` and `version.rs`. `couple --base origin/main --head HEAD` passes locally and in the self-governance job.
2. **`authority` with empty edge maps after a sugar rewrite: intended, D-10.** The class is decided on raw values so an edit to an edge key is never read as nothing (D-4), and the item lists are computed on the parsed edges so a `paths:` or full-scope `supersedes` respelling does not invent an added or removed edge. Empty maps on an `authority` entry are the report saying exactly that: the key's text moved and no parsed edge did. `baseOwners` and `headOwners` are still carried.
3. **Trailing newline in the body: reported, as `unknown`.** `without_verification_section`, `split_frontmatter` and the plan parser all read `str::lines()`, so an EOF newline or CRLF-only change leaves no rule placing it. Such a `spec.md` then takes 3.3's last row, `unknown`, which sets `priorPolicy.required` (D-8). It is not reported as an unchanged file.
4. **`resolve_commit`:** agreed, no issue.
5. **Full-tree export: required, not incidental.** The classifier compares the base's committed index with its whole tree under spec 086 (3.2), indexes the whole head tree in memory for head ownership (D-5), and resolves symbol, section and directory units against source outside the changed set. A selective export would make both of those read a partial tree. Measured on this repository's own branch: 19 paths in 0.5 s.

No code change from this pass.

PR #199, opened 2026-09-15
docs(097): key report omission to the scope and bind ignore rules to untracked files

bartekus:
Two wording corrections to draft 097 from the pre-build review, before its build.

- **Ignore rules bind only untracked files (§3.6).** The prescribed `git ls-files -z --cached --others --exclude-standard` lists tracked files even when `.gitignore` matches them, because `--exclude-standard` filters only `--others`. The command stands; the text now retains tracked files and excludes only untracked ignored ones. §3.8 gains CLI cases for tracked-ignored (retained), untracked-ignored (excluded), unstaged additions (included) and missing files (dropped).
- **Omission is keyed to the configured scope (§3.5, §3.8).** Both `declaredScopeFiles` and `enumeration` are omitted only when `governed_scope` is empty. A set scope that matches nothing emits `declaredScopeFiles: []` with its `enumeration`, which is exactly the case where supplied-empty must be told apart from a walk. The unset-key verification line now asserts absence instead of accepting an empty array.

Recorded as 097 D-6, with a paragraph in note 05 §7.1. `index coverage --paths-from` is kept. Build order unchanged: 092, 095, 096, 094, 093, 087, 097.

Gate: compile, index, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` all pass. Spec text only; no code.


bartekus:
ai-review's coupling question answered: 097 is `implementation: pending` and none of its surface exists yet. `git grep 'governed_scope\|declaredScopeFiles\|declared_scope' -- crates/` on this branch returns nothing, so no test in `tests/coverage.rs` encodes the old §3.5 or §3.6 wording, and the four new §3.8 CLI cases are requirements for 097's build PR. No change.

PR #200, opened 2026-09-15
feat(092): judge mode-only and binary changes at the coupling gate

bartekus:
Builds draft spec 092 (stays `draft`; ratify PR follows the merge).

## What changes

`couple` registered a path only from the `---`/`+++` headers of `git diff -U0`, and git prints neither for a mode-only change or a binary change. Both arrived as no change: an owned binary could change without its spec and `C-001` stayed silent, and a mode flip on an unclaimed source file never reached `C-002`.

`cmd_couple.rs::build_diff_input` now unions the parsed diff with `git diff --name-status -z --no-renames` over the same `base...head` range (§3.1):

- a path the parser missed enters as a whole-file change (no hunks), `deleted` exactly when the status letter is `D` (§3.2);
- a path the parser registered keeps its spans and deletion verdict;
- `changed_path_names` (088's `delta` helper) is now expressed through the shared `changed_path_statuses`, same flags (`core.quotepath=false`, `--no-renames`, `--end-of-options`, `-z`), so `couple` and `delta` cannot disagree (§3.3).

No core change, no schema change, no exit-code mapping change. `--paths-from` is untouched (§3.5).

## Evidence

- `tests/couple.rs`, through the binary over real repositories: mode flip on a claimed path is `C-001`; binary edit on a claimed path is `C-001`; a binary delete is judged (`checkedPaths` 1) and is not `C-002`, with the add of the same unclaimed binary as the `C-002` control; a mode flip with its owning `spec.md` clears. **All four fail against the pre-092 adapter** (probe: `git stash push -- crates/spec-spine-cli/src/cmd_couple.rs`, 4 failed).
- A text edit plus mode flip on one path is judged once (regression guard for the union).
- Adapter unit tests: `--name-status -z` parsing (newline in a path, rename pair, status with no path is an error), union ordering/spans, and span survival over a real repository.
- `spec-spine verify 092`: 6/6 pass. Workspace tests 701 passed, clippy and fmt clean, full governance gate green.

## Decision recorded

**D-3**: §3.7 case 5 ("a text change still reports its hunk spans") cannot be observed through the binary. `index.rs` seeds a whole-file implementing path for every owning unit's file, and `owners_for_path` reads that as whole-file ownership, so no `couple` verdict depends on spans for an index-resolved unit (measured: a section-owned Makefile edited only outside the owned target is refused with no mode change involved). The span half is asserted at the adapter over a real repository; the binary asserts the judged-once half. That section granularity is inert at the gate is a separate finding in spec 004's territory and is not changed here.


bartekus:
ai-review, both items answered, no change:

1. **R/C branch in `parse_name_status_z`.** Kept deliberately. Under `-z` a rename or copy record carries two NUL-terminated paths, so a parser that did not consume both would misread every later record as status/path pairs out of step. `--no-renames` being dropped is not masked: it is pinned by 092's own verification line `grep -qF -- '--no-renames' crates/spec-spine-cli/src/cmd_couple.rs` (§4), and a move would still reach the gate as its destination path rather than vanish.
2. **Range semantics.** Confirmed in source: `run_git_diff` passes `format!("{base}...{head}")` (`cmd_couple.rs`, the `.arg` after `--end-of-options`), and `build_diff_input` passes the same three-dot operand to `--name-status`. `delta`'s two operands are `merge_base(base, head)` and `head`, which is the same range spelled two-dot.

PR #202, opened 2026-09-15
feat(095): keep the freshness verdict when a committed shard will not parse

bartekus:
Builds draft spec 095 (stays `draft`; ratify PR follows the merge).

## What changes

`index check` and `check` computed the index freshness verdict (which, since spec 086, already calls an unparseable stray `orphaned` and a corrupted expected shard `modified`) and then called `diagnostics::committed_counts(..)?`, which parsed every shard. The `?` discarded the verdict: exit 3, parse error, where `couple` and `index coverage` exit 2 and name the orphan.

- `committed_counts` is best-effort (§3.3): a shard that will not deserialize, or carries a foreign MAJOR, is skipped and named in `DiagnosticCounts::unreadable` (`#[serde(skip)]`). It errors only on an index never built.
- `verdict_tally` (`lib.rs`) is the one function `check_report`, `check_freshness_json`, `index check` and `check` call after the verdict (§3.1, §3.4).
- `IndexCheckReport.skippedShards`: `#[serde(default, skip_serializing_if = "is_zero")]`, so a fresh tree's payload gains no member. `VERDICT_SCHEMA_VERSION` does not move.
- Prose: each skipped file's drift line gets `(unreadable: not counted in the diagnostics tally)`.
- Consumers keep refusing (§3.2): `index owner` at its freshness guard, `index render`/`orphans`/`diagnostics` at the reader.

## Evidence

At the verbs (`tests/cli.rs`), per §3.6:

| Case | Tree | Result |
|---|---|---|
| 1 | `{"nope": 1}` in `by-spec/` | exit 2, orphan named, `skippedShards: 1`, facade == CLI payload |
| 2 | valid shard copied to a new name | exit 2, nothing skipped |
| 3 | `[]` in `by-package/` | exit 2, orphan named, skipped 1 |
| 4 | expected shard overwritten with garbage | exit 2, named `modified`, not `orphaned` |
| 5 | fresh | exit 0, index payload keys exactly `diagnostics, fresh, unwitnessed` |
| 6 | case 1's tree | `index owner` exit 2, `index render` exit 3 |

**Fail-first:** with the four source files stashed, cases 1, 3 and 4 fail; the facade test added to 086's `index_body.rs` fails too (stubbing the two lines that name new fields). Cases 2, 5, 6 are the must-keep-working half.

`spec-spine verify 095`: 7/7. Workspace tests 708 passed; clippy, fmt, `--no-default-features` core tests clean; governance gate green.

## Decisions recorded

- **D-5:** §3.6 case 6 said `index owner` "still refuses at exit 3" on case 1's tree. Measured at 0.19.0 before any change it exits **2** (its freshness guard names the orphan), as §3.2's own list says. The case asserts the consumer half as it is: `index owner` at 2, and `index render`, which has no guard, at 3.
- **D-6:** the count is the payload member; the names ride unserialized for prose; `committed_counts` keeps its signature (spec 050's tests call it); a foreign MAJOR is skipped by the tally because 086 D-5's refusal already runs first in the freshness comparison.


bartekus:
ai-review notes answered, no change; none names a defect against today's behavior:

- **Drift-line format coupling in `annotate_unreadable`.** The fallback is deliberate: a name no drift line shows gets its own `unreadable <file>` line, so a future format change degrades to a separate line and never to an unnamed skip (§3.3's "MUST NOT be silent"). The line format is spec 031 §3.3's contract, pinned by 086's tests.
- **`verdict_tally`'s `unwrap_or_default`.** Its doc comment states the one error and why the default is not a verdict. The race named (index directory removed between the two reads) yields zero counts beside the verdict already computed, which is §3.1's ordering: the verdict outranks the tally.
- **`skipped_shards` in the fresh arm.** Computed the same way in both arms on purpose, so the report states what the tally saw rather than encoding an invariant the constructor cannot check; `skip_serializing_if` keeps the fresh payload byte-identical (§3.6 case 5 asserts the key set).
- **`committed_counts` semantics.** Documented on the function itself ("Best-effort since spec 095"), and D-6 records why the signature is kept. The consumers that must refuse a corrupt ledger read through `read_committed_index_shards`, unchanged, and §3.6 case 6 asserts it.

PR #203, opened 2026-09-15
feat(096): name the path framing where registry show prints contentHash

bartekus:
Builds draft spec 096 (stays `draft`; ratify PR follows the merge). Carries the wave's `amends` edge on approved spec 055, authored in 096 per spec 040; 055 is not edited.

## What changes

`registry show` printed `contentHash: <h>  (sha256 of this spec.md)`, and 055 §3.4 says the same. The value is, and always was, the registry `shardHash`: `SHA-256(<repo-relative POSIX path> 0x00 <normalized bytes>)`. The plain digest of the normalized bytes is a different value, which `attest --spec` reports as `specSourceHash`.

- `cmd_registry.rs`: the line now reads `(sha256 over path + NUL + normalized spec.md bytes; not the bare file digest)` (§3.2). No value, DTO, artifact or schema changes.
- Docs (§3.5): `docs/api.md` names both digests and which verb reports each; `website/docs/cli/registry.md` states the construction beside the field; `docs/authority-evidence.md` §4 records the discrepancy as closed. None is an `extra_hashed_inputs` match.
- `crates/spec-spine-cli/Cargo.toml`: `sha2 = { workspace = true }` as a dev-dependency (D-4), carried by an `extends` edge. `Cargo.lock` gains one dependency line; no new package.

## Evidence

`content_hash_is_path_framed_and_spec_source_hash_is_not` (§3.4), over a spec written with a BOM and CRLF so normalization is covered too:
1. `registry show --json` `contentHash` == SHA-256(path, NUL, normalized bytes), computed in the test;
2. != SHA-256(normalized bytes);
3. `attest --spec --json` `specSourceHash` == SHA-256(normalized bytes);
4. the prose line lacks `sha256 of this spec.md` and names `path` and `NUL`.

**Fail-first:** with `cmd_registry.rs` stashed, (4) fails on the shipped line. `spec-spine verify 096`: 7/7, including the Python recomputation lines and both prose lines. Workspace tests 702 passed; clippy and fmt clean; governance gate green.

## Decision recorded

**D-4:** the pin computes SHA-256 itself via a `sha2` dev-dependency. The core's `hash` module is private, and a pin borrowing the tool's own function would agree by construction; `shasum`/`sha256sum` availability differs by platform.


bartekus:
ai-review notes answered, no change:

- **Normalization fixture.** The spec text comes from `verdict_fixture`, which writes a Rust string literal with `\n` line endings and nothing else, so the file cannot hold `\r\n` before the test adds it. `.gitattributes` does not reach a tempdir file.
- **Doc ownership.** Neither doc has an owner: `spec-spine index owner website/docs/cli/registry.md` prints `(no spec owns this path)`, and `docs/` is on the built-in bypass floor (`couple.rs::DEFAULT_BYPASS_PREFIXES`). `couple` passed locally and in self-governance. `references` is non-owning (spec 034) and the edit needs no edge.

PR #205, opened 2026-09-15
feat(094): declare the comment-header claim window and report near misses

bartekus:
Builds draft spec 094 (stays `draft`; ratify PR follows the merge). Unblocks 097.

## What changes

A `// Spec:` header claims from `content.lines().take(16)`, a bound no document recorded. Three failures were silent: a header on line 17, a header naming a spec not in the corpus (which also shadows a valid header below it), and `//! Spec:`.

- `index.rs`: `COMMENT_HEADER_CLAIM_WINDOW = 16` (value unchanged, §3.1) and `COMMENT_HEADER_REPORT_WINDOW = 64` (§3.6). The recognizer's steps 1-3 are one function (`header_attempt`) and the file walk is one function (`header_scan_files`), both shared by the claim scan and the new near-miss scan. The claim scan's behavior is unchanged.
- `CoverageReport.nearMissHeaders` (§3.4): `{path, line, reason, specId?}`, reasons `outside-window`, `unknown-spec`, `doc-comment-marker`, sorted by path then line, omitted when empty. Computed on read; no shard changes. The classifier and `--fail-on-untraced` do not read it (§3.5).
- Prose `index coverage` prints a `near-miss comment headers` block when non-empty.
- `website/docs/cli/index.md` states the window and recognizer (§3.7). It also corrects the `index check` exit-3 line, which still said an unparseable shard exits 3; spec 095 (#202) made that drift.

This corpus reports **no** near miss, as §3.6 measured.

## Evidence

- `tests/index.rs` (§3.8): line 16 claims, line 17 does not; `//!` does not claim; `#` claims in `.sh`; an unknown reference stops the scan and shadows a valid header; one regression case per row of §3.2's table. **All 52 index tests also pass against the pre-094 scanner** (constant assertion removed for the probe), so no claim moved.
- `tests/coverage.rs`: each reason with path, line, reason and spec id through `coverage()` and `coverage_json`; the claiming file and a header on line 65 produce none; classification identical with and without the header lines; report-window edge at 64.
- `spec-spine verify 094`: 7/7; its three grep lines return 0 matches on `origin/main`. Workspace tests 716 passed, `--no-default-features` core 496 passed, clippy and fmt clean, governance gate green.

## Decision recorded

**D-4:** §3.3 left three things open. The window scan reports up to and including the first claim attempt (where the claim scan stops); `outside-window` reports every resolving header on lines 17-64; and both scans share one recognizer and one enumeration.


bartekus:
ai-review, three findings answered, no change:

1. **`index.rs` absent from `extends`.** It is present: 094's first `extends` entry is `{ spec: "004-codebase-index", unit: "crates/spec-spine-core/src/index.rs", nature: additive }`, which predates this PR; the diff shows only the `types/src/lib.rs` line this PR added. `couple` passed locally and in self-governance.
2. **Second read pass in `coverage()`.** Accepted cost. `coverage()` already runs a full `index()` in its freshness guard (spec 086) and a second walk in `enumerate_source_files`; the near-miss scan is one more pass over the same small source set on a read verb, and threading the index's intermediate state out would reach into the staleness gate, which spec 050 §3.5 deliberately leaves alone for the same trade.
3. **`unknown-spec` omits the reference text.** §3.3 fixes the record as path, line, reason and the spec id when it resolves; a raw-reference member would be an unspecified addition to the DTO. The `path:line` pair points at the text. A reference member is a reasonable additive follow-up, not part of this build.

PR #208, opened 2026-09-15
feat(087): emit an authority snapshot that says what it read

bartekus:
Builds draft spec 087 (stays `draft`; ratify PR follows the merge). Authorized for build 2026-09-15 with the piece rule (§3.3.1), the `d` kind (D-7) and the unavailable join hash (D-8) settled.

## What changes

- **`attest --snapshot [--sign --key] [--json]`** writes `<derived>/attestation/snapshot.json` (and `snapshot.sig`), an `AuthoritySnapshot` on its own axis, `SNAPSHOT_SCHEMA_VERSION = 0.1.0`. Exit 0 whenever a payload is written. Combined with `--spec` or `--with-coupling`: exit 3, "cannot combine".
- **`verify-attestation --snapshot --recompute | --signature`** under spec 085's rules (stored bytes, MAJOR gate, strict parse); a mismatch names the member that moved (`specs[<id>]`, `committed.index`, ...).
- **Facades:** `attest_snapshot_json`, `verify_snapshot_attestation_json`.
- **`frame/1`** (`core/src/snapshot.rs`): SHA-256 over `"spec-spine/frame/1" 0x00` then, per path-sorted deduplicated piece, a kind byte (`t`/`b`/`l`/`d`), u64-BE-length-prefixed path and content.
- **Payload** (§3.1): config digest; schemas; corpus (spec count, `inputsManifestHash`, `registryHash`); committed registry/index `{files, hash, matchesRecompute}`; governance inputs by path; verdicts (compile counts, lint, resolution counts, ownership counts, unwitnessed pair); per spec `status`, `implementation`, `territoryDigest`, and `specAttestationHash` or `specAttestationUnavailable: "non-utf8-direct-claim"`; exclusions.
- **`attest_spec`** is split into crate-visible helpers (`spec_units`, `build_spec_attestation`) so one compile/index/lint serves every spec. Its output bytes and error text are unchanged (§3.7).
- Docs: `docs/authority-evidence.md` gains the snapshot's digest table and what it does and does not establish (§3.6); `docs/schema-versioning.md` gains the axis row (a hashed input, so every index shard is regenerated here).

## Evidence

- `tests/snapshot.rs` (20): length framing, all four kind bytes, `b` vs normalized `t`, path dedup and no digest for an empty set; the piece rule (overlapping claim not doubled, all-unresolved has no digest, empty directory vs empty file, empty claims distinguished by place, whole-file binding, symlink unfollowed); non-UTF-8 direct claim keeps `territoryDigest` with the stated reason while `attest_spec` still errors; join hash equals `attest_spec`'s; determinism; absent trees/config; `matchesRecompute` moves while `registryHash` does not; governance inputs; recompute match, named mismatch and version mismatch; non-canonical stored bytes; unknown member refused.
- `tests/cli.rs`: write, envelope == facade, seal + both verify modes, facade verify, named mismatch after an edit; both scope refusals and the verify-side one; snapshot join hash == `attest --spec --json`'s hash.
- `spec-spine verify 087`: **22/22**, including the `d`-kind, dedup, unresolved, separation and unavailable-join lines.
- Workspace tests 752 passed; `--no-default-features` core clean; clippy and fmt clean; governance gate green.

## Decision recorded (D-9)

Seven choices the text left open. The one to review: **every empty directory the walk reaches is a `d` piece**, not only a claimed one. Under spec 083's walk a nested empty directory contributed nothing, so the Verification block's empty-directory line passed without ever producing a `d` piece. `index::empty_territory_dirs` yields leaf empty or wholly pruned directories and shares the walk's pruning predicate, so the file set is still exactly 083's. The others: a directly claimed symlink is an `l` piece; `governanceInputs.hash` is absent with no paths; the index `files` count includes the slices sidecar while `matchesRecompute` compares shard directories; verdicts come from the recompute so a stale tree still yields a record; `attest_spec` is split rather than duplicated; `verify-attestation --snapshot --spec` is refused.


bartekus:
ai-review answered; two taken, two declined, one noted:

- **Duplicated `BLOCKING_RESOLVER_CODES`.** Taken: the snapshot now imports `attest.rs`'s list rather than carrying a third copy, so `resolution.blocking` and spec 023's `couple.ok` cannot count different codes.
- **The `"specs (order)"` fallback.** Taken: it now reads "specs order, or a member this build does not name", which is what the guard actually covers.
- **`extends` for `main.rs`, `core/src/lib.rs`, `types/src/lib.rs`.** All three are declared in 087's frontmatter from its filing (001 for the first two, 000 for the third), and `couple` passes locally and in self-governance.
- **`empty_territory_dirs` skipping an entry whose metadata cannot be read.** Deliberate and consistent: `walk_territory` skips such an entry too (`let Ok(meta) = symlink_metadata(..) else { continue }`), so the two functions agree about what a subtree contains, which is what §3.3.1 requires of them. Counting it would make the snapshot's piece set disagree with `attest --spec`'s.
- **Recompute cost.** Inherent and shared with the other two scopes: a recompute verifies by recomputing.

Rebased onto main (the 093 and 094 ratify merges touched the same generated shards). Re-gated after the rebase: 752 workspace tests, clippy and fmt clean, `verify 087` 22/22.

bartekus:
Second pass answered:

- **`empty_territory_dirs` emitting a `d` piece for an unreadable directory.** It does not: `let Ok(entries) = fs::read_dir(dir) else { return; };` returns **before** the `if children == 0 { out.push(..) }` at the end of `visit`, so an unreadable directory yields no piece at all, exactly as `walk_territory` yields no entry for it. Pushed b5d2ab7, which adds a comment saying so; no behavior change.
- **`spec-spine.toml` read twice.** Deliberate: `config` and `governanceInputs` are two members answering two questions (what configuration was read; what feeds the index's global scalar), and the second is a list a corpus can change independently. One extra read of one small file on an on-demand verb.
- **`docs/authority-evidence.md` ownership.** `spec-spine index owner docs/authority-evidence.md` prints `(no spec owns this path)`, and `docs/` is on the built-in bypass floor; `couple` passes here and in self-governance.
- **Version mismatch hiding a content delta.** Consistent with both other scopes by design (spec 023 FR-005): a recompute is only meaningful under the same tool version, so the version answer is the honest one.

bartekus:
Third pass, no new defect; each item settles against the tree:

- **`verify_attestation.rs` not in `extends`.** It is: `{ spec: "023-ledger-seal", unit: "crates/spec-spine-cli/src/verify_attestation.rs", nature: additive }` has been in 087's frontmatter since it was filed, beside the `cmd_attest.rs` edge. `couple` passes locally and in the self-governance job.
- **`docs/authority-evidence.md`.** `index owner` reports no owner, and `docs/` is on the built-in bypass floor. `references: context` is the correct annotation for a document this spec cites and does not govern; 087 §3.6 requires the update, and spec 034 makes `references` non-owning on purpose.
- **A missing `authority-snapshot.schema.json`.** There is no schema file for either attestation either: `crates/spec-spine-types/schemas/` holds `build-meta`, `codebase-index`, `codebase-index-spec-shard`, `codebase-index-package-shard`, `registry` and `registry-spec-shard` and nothing else. The conformance test validates the **committed artifacts**; an attestation and a snapshot are on-demand records, gitignored, and verified by recompute (spec 085), which is a stronger check than a shape schema.
- **`symlink_metadata` skipping an entry.** Answered on the first pass: `walk_territory` skips such an entry identically, and the two functions agreeing is what §3.3.1 requires; counting it would make the piece set disagree with `attest --spec`'s file set.

Merging: three passes, and this one names no code path the earlier two did not.

PR #211, opened 2026-09-15
feat(097): declare the governed scope instead of inferring it

bartekus:
Builds draft spec 097 (stays `draft`; ratify PR follows the merge). Last of design note 05's wave A. Its dependency 094 is complete and approved.

## What changes

The ratchet's universe had four conjuncts and nobody declared the third: a file was governable because a manifest happened to sit above it. Measured in §1.2: 572 tracked files, 89 visible.

- **`[coverage] governed_scope` / `governed_scope_exclusions`** (§3.2): glob patterns over repo-relative paths, matched with `extra_hashed_inputs`' own matcher so the `dir/**` trap behaves identically (the `init` default says so). Exclusions carve out of the scope's addition only.
- **Universe** (§3.3): a scoped path joins the coverage universe whatever its extension and wherever it sits, and so joins `C-002`. `resolver_exclusions` and the bypass verdict still win; spec 009's claim precedence is untouched; the state root is unreachable. The claim scanner is **not** widened, so §1.2's seven inert headers stay inert (§3.4).
- **Report** (§3.5): `declaredScopeFiles` and `enumeration`, present exactly when the scope is configured, so a set scope matching nothing still emits `[]` with its provenance. Those files count toward the totals and toward no package; the prose gives them their own `declared scope (<enumeration>)` line.
- **Enumeration** (§3.6): the CLI supplies it, because the core has no git. `git ls-files -z --cached --others --exclude-standard` minus paths the tree no longer holds (`tracked`), `index coverage --paths-from FILE` (`supplied`), or a root walk skipping `.git/` and the state root for a library caller that supplies none (`walk`). A git failure is **exit 3 naming `--paths-from`**, never a silent walk. `coverage_inventory_json` is the facade half.
- **`config show`** prints both keys (§3.7); `init` writes the commented default.

**Empty by default**: with the key unset no verdict, count or classification moves, no git runs, and neither member is emitted. The two additive outputs §3.1 names (the `config show` keys, the `init` comment) are the stated exceptions.

## Evidence

- `core/tests/coverage.rs`: unset and `[]` byte-identical and carrying neither member; a file outside every package and one with no source extension entering; exclusions removing only what the scope added; a resolver exclusion and a bypassed path staying out despite matching; a set scope matching nothing keeping both members; absent vs supplied-empty as two different reports, and `tracked`/`supplied` provenance carried; the walk skipping `.git/`, the state root and `target/`.
- `core/tests/couple.rs`: an unclaimed scoped file is `C-002` naming no floor; the claimed one with its spec is clean; an unclaimed file under a built-in bypass prefix stays bypassed while the same prefix with an explicit unit claim is governed as spec 009 makes it (`C-001` without its spec, clean with it); with no scope the same file is not asked about.
- `cli/tests/cli.rs`: `--paths-from` as the supplied inventory, equal to the facade's report for the same inventory, and an empty list as an answer; the git failure at exit 3; the four git membership cases (tracked-ignored retained, untracked-ignored excluded, unstaged addition included, missing dropped); `config show` and a scaffolded `spec-spine.toml`.
- `spec-spine verify 097`: **12/12**. Workspace tests 766 passed, `--no-default-features` core 536, clippy and fmt clean, governance gate green.

## Decision recorded (D-7)

Five choices the text left open: glob membership through `shard::glob_files` (so the trap matches `extra_hashed_inputs`), no enumeration for `couple` (every path it judges is already in the diff), no git while the scope is empty, scope-only files counted in the totals and in no package, and a new facade function rather than a changed `coverage_json` signature.

## Not in this PR

Turning the scope on in this repository (§4): that is a separate change which must first give every named file an owner, and it edits `spec-spine.toml`, which restales every shard.


bartekus:
ai-review: no defect named; three notes answered, no change:

- **Double walk in the walk path.** Deliberate. §3.2 requires the `extra_hashed_inputs` glob semantics, trap included, and those are the semantics of globbing the tree (a string matcher would let `dir/**` match files). The inventory walk answers a different question, which is which files the scope may match at all. Both are read-verb costs on a verb that already runs a full `index()` for its freshness guard.
- **`Vec::contains` in `render_coverage`.** Prose rendering over a report a person reads; the lists it scans are the coverage debt of one corpus. Left as the surrounding code reads.
- **`symlink_metadata` per tracked path.** The stat is what makes the "missing file is dropped" case true (§3.6), and the alternative is a second git invocation. One stat per tracked file, only while a governed scope is configured.
- **`establishes` for `cmd_index.rs`, `cmd_config.rs`, `scaffold.rs`, core `lib.rs`.** Each is covered by an `extends` edge in 097's frontmatter (032, 054, 006 and 001 respectively), which is the correct edge: 097 establishes no new file. `couple` passes locally and in the self-governance job.

PR #212, opened 2026-09-15
chore(087): ratify (draft -> approved)

bartekus:
## Summary

Flips spec 087 (`an authority snapshot says what it read`) from `draft` to `approved`. Built and merged in #208 (squash `4799b64`); its `## Verification` block passes on that sha: **22 commands, all exit 0**. Only the lifecycle key changes.

## What becomes governing

`attest --snapshot` and `verify-attestation --snapshot`, the `AuthoritySnapshot` payload on its own axis (`SNAPSHOT_SCHEMA_VERSION` 0.1.0), and **`frame/1`**: the length-prefixed, kind-tagged, path-deduplicated digest that new records use from here (design note 05's wave C is gated on exactly this). The existing constructions are untouched: `CorpusAttestation` and `SpecAttestation` are byte-identical, `attest --spec` still refuses a directly claimed non-UTF-8 file, and `specAttestationHash` is historical evidence rather than a content binding.

D-9 records the seven build-time choices; the load-bearing one is that every empty or wholly pruned directory the territory walk reaches is a `d` piece, without which the spec's own empty-directory verification line passed while never exercising that piece kind. Three ai-review passes on #208 were answered in-thread; two findings were taken (one shared blocking-code list, a clearer fallback difference).

## Testing

`spec-spine verify 087` on `4799b64`: passed (22 commands). Gate on this branch: `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` all pass.


bartekus:
No attestation artifact is required for a lifecycle flip: attestations (specs 023, 042, 087) are on-demand, gitignored records, not committed evidence, and spec 049's acceptance is the spec's own `## Verification` block, which passes 22/22 on `4799b64` as the body cites. The draft-through-build cadence is this repository's own (AGENTS.md, "Working the backlog"): a spec is filed `draft`, built, then ratified, which is why `implementation: complete` precedes `status: approved`.

Issue #216, opened 2026-09-16
Freshness hooks still print a remedy the verdict no longer names (Stop + SessionStart)


bartekus:
Spec 098 corrects what `check` and `index check` *say* when a spec claims a unit
that does not resolve: the refusal is reported as an unresolved claim rather
than as staleness, and it states that regenerating will not clear it. 098 §3.5
and §3.5.1 deliberately leave the two kit hooks alone, and this issue is the
carry-forward they name, so the misleading-remediation problem is not reported
as closed on 098's strength.

**What remains**

- `Stop` hook (`kit/settings.json`, `.claude/settings.json`): runs
  `"$sc" check >/dev/null 2>&1` and prints `[freshness] STALE: run
  \`spec-spine compile\` and \`index\`` on *any* nonzero. It discards the output
  098 corrects, and it prints the same remedy for exit 1 (validation, refused
  unresolved unit) and exit 3 (I/O, parse, schema, config) as well.
- `SessionStart` hook: matches `codebase-index: STALE` in `check`'s output.
  After 098 that match still fires for the stale and mixed cases, which is
  correct. In the blocking-only case it no longer fires and the hook falls back
  to `unknown (check exit 2)`: uninformative, but no longer false. 098 §3.5
  records that as a deliberate trade rather than a fix.

**Why it was not done in 098**

- It needs a policy decision (does the `Stop` hook refuse, or advise?), which is
  design note 06 §3.9's open question H-6. A message fix has no standing to take
  it.
- Both files are hashed governance inputs (`[index] extra_hashed_inputs`), so
  editing either restales every shard in the corpus.
- The hook exists in two pinned-identical copies plus the generated
  `kit_embedded.rs`, and its blast radius reaches every adopter who re-copies
  the kit.

**Scope when it is filed**: a spec of its own, covering both hooks, the exit-code
classes each should distinguish, and the adopter upgrade note.

Related and deliberately separate: whether an `I-004` refusal should exit `1`
rather than `2` (note 05 R-4) is a contract change needing an `amends` edge on
spec 086 §3.1; 098 §3.1 holds the current codes.

bartekus:
A third copy of the `SessionStart` match exists outside the two `settings.json` files: `.codex/hooks.json` runs the same `check` read and matches `codebase-index: STALE` the same way, so it falls to `unknown (check exit $c)` in the blocking-only case for the same reason. No spec owns that file. Whatever decision H-6 takes has to reach all three copies.

PR #217, opened 2026-09-16
chore(098): ratify (draft -> approved)

bartekus:
## Summary

Flips spec 098 (`a blocking claim is not a stale shard`) from `draft` to
`approved`. Built and merged in #215 (squash `63ae87a`); its `## Verification`
block passes on `ae466ec`: **58 commands, all exit 0**. Only the lifecycle key
changes.

With this, the corpus reads **99 specs, 99 approved, 0 draft**.

## What becomes governing

`check` and `index check` distinguish two refusals they previously folded into
one. A spec claiming a unit that does not resolve is reported as an unresolved
claim, not as staleness, and the report says that regenerating the index will
not clear it. Each line names the diagnostic code, the owning spec and the
unit; when the owning spec declares `implementation: complete`, the report says
the spec and the tree disagree about what exists. The mixed case reports both
halves and attributes regeneration to the stale half alone.

The exit codes do not move: a blocking diagnostic, a stale tree, and the two
together all still exit `2`, and `--fail-on-unresolved` keeps its own meaning
over the warning tier. The `--json` envelope does not move either: the
`blocking-diagnostics <shard>` line survives in `freshness()`, one line per
shard, so a machine caller reads exactly what it read before. Whether an
`I-004` refusal should exit `1` is a separate contract change needing an
`amends` edge on spec 086, and 098 3.1 deliberately holds the current codes.

The report offers no way out of the refusal. It does not suggest narrowing the
claim, and it does not suggest `planned: true`, which beneath
`implementation: complete` is spec 076 3.3's `L-011`.

## Not fixed here

The two kit freshness hooks (`Stop`, `SessionStart`) still print a remedy the
verdict no longer names. 098 3.5 leaves them alone deliberately: the fix needs
design note 06 3.9's H-6 policy decision (does the `Stop` hook refuse or
advise?), it edits hashed governance inputs, and it reaches every adopter who
re-copies the kit. Carried forward as #216.

## Testing

`spec-spine verify 098` on `ae466ec`: passed, 58 commands. Gate on this branch:
`lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` all exit 0, with the two regenerated
shards committed alongside the flip.


bartekus:
On the status-promotion completeness note, recorded here because the PR is
merged and the question is a fair one to have asked.

**The premise does not hold.** Neither spec requires what the note attributes
to them, and both explicitly disclaim it.

- **041** governs the *in-flight predicate*, keyed on `implementation`, not on
  `status`. Its 3.3 is titled "What this does not do" and says it verifies
  **existence, not behavior**; its 4 repeats that no gate reads English and
  "one that claimed to would be worse than none". It says nothing about the
  `draft -> approved` transition.
- **049** is the `verify` verb's grammar and the core/CLI seam. Its 4 lists
  "Verifying that code matches a spec's prose" as out of scope, unchanged from
  041 4 and 044 4.

So there is no requirement that a spec reaching `approved` has its behaviors
witnessed, and no attestation artifact is owed on a ratify diff.

**The conditional resolves to: CI does not enforce it, the loop does.**
`verify` is deliberately outside the gate chain, because it is the one verb
that *executes* rather than reads and the chain runs against PR branches whose
contents are in the general case a stranger's. `grep verify .github/workflows/ci.yml`
returns nothing. What orders it is `AGENTS.md` step 6: `/verify` runs the
block, `/ship` sets `implementation: complete` only once it holds, and the
ratify PR comes after that merge. Enforcement is procedural, by construction
of the cadence.

**It was satisfied here, and the diff is the wrong place to look for it.**
`spec-spine verify 098` was re-run on the merged sha `ae466ec`, not on the
build branch: passed, 58 commands. That is cited in this PR's body.

**The existence half is mechanically enforced, and did run on this PR.** 041's
table makes `draft` + `complete` with an unresolved owning unit a hard `I-0xx`
error rather than a `W-001` warning, so a promotion whose claimed units do not
resolve fails `check --fail-on-unresolved --fail-on-warn` in the
self-governance job. That was green here.

The residue worth keeping: nothing mechanically blocks a ratify PR whose
`## Verification` block fails, only the cadence does. Design note 05 9.3
already disposes of the adjacent question (`draft` + `implementation: complete`
is intentional in this corpus and a lint refusing it would refuse this
repository's normal state). Gating ratification on verify would be a new
decision, not a bug fix.

PR #218, opened 2026-09-16
docs(098): stop listing blocking-diagnostics as a prose drift class

bartekus:
## Summary

`website/docs/cli/index.md` listed `blocking-diagnostics` as a fourth per-shard
drift class `index check` prints, next to `modified`, `missing` and `orphaned`.
Spec 098 (#215) moved it. The page had not caught up.

## What the verbs actually do now

- **Prose.** `cmd_index.rs` and `cmd_check.rs` print the stale half through
  `stale_verdict()`, which carries only the three shard classes, and then report
  an unresolved claim separately under `<subject>: UNRESOLVED CLAIM: N
  unresolved claim(s) over M spec(s), which is not staleness`, one line per
  diagnostic, closing with the statement that regenerating does not clear it.
- **`--json` and the library verdict.** Unchanged. `IndexFreshnessReport::
  freshness()` still folds the blocking half in as `blocking-diagnostics
  <shard>`, one line per shard rather than one per diagnostic.
- **Exit code.** `2` either way, per 098 3.1.

So the class is not gone; it is no longer something the prose prints. The page
now says which surface it belongs to, and documents the unresolved-claim
reporting the prose gained in its place, including the
`implementation: complete` contradiction line and the mixed-case attribution of
regeneration to the stale half alone.

The list intro changes from "Each drifted shard is named with its class" to
"Each stale shard", since after 098 that is what those three classes cover.

## Scope

One file, documentation only, inside the bypass floor and outside
`[index] extra_hashed_inputs`, so no shard is restaled: the gate leaves the
tree with only `website/docs/cli/index.md` modified.

## Testing

Gate on this branch: `compile`, `index`, `check --fail-on-unresolved
--fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` all exit 0. Every claim above was read
back out of `crates/spec-spine-core/src/index.rs`,
`crates/spec-spine-cli/src/cmd_check.rs` and `cmd_index.rs` rather than from
the spec text.

Stacked behind #217 only by subject; the two do not touch a common file.


bartekus:
Disposition of the automated review, with the evidence for each.

**1. `blocking-diagnostics` discoverability — accepted, fixed in `840906d`.**
Fair reading. The paragraph did say the class survives, but nothing in the
bullet list pointed there, so a skim of the three prose classes reads as a
removal. The class is back in the list, marked machine surface only, with the
paragraph carrying the detail.

**2. The output-format string read as a contract — accepted, fixed in `840906d`.**
This one was right on substance, and checking it changed the text. Spec 098
3.3 fixes the *content* of the report (classed as an unresolved claim, one
line per diagnostic naming code, owning spec and unit, the statement that
regenerating does not clear it, the pointer to `index diagnostics` when
capped), not the wording. 098's `## Verification` block bears that out: it
asserts `I-004`, `001-missing-territory`, `src/nothing.rs`, `regenerat` and
the mixed-case attribution, and asserts nothing about `UNRESOLVED CLAIM:` or
`which is not staleness`. The page now states the classed facts as the
contract, marks the phrasing illustrative, and tells machine callers to match
on the diagnostic code rather than the sentence.

**3. Granularity cross-reference — no change, as the review itself proposed.**
Raised as a note with no finding. The envelope link carries the shape and the
per-shard vs per-diagnostic distinction is stated where it applies.

One thing the review did not flag, found while acting on it: the fix for (2)
first shipped a link to `#index-diagnostics`, an anchor that does not exist.
`index diagnostics` has no section on this page or anywhere under
`website/docs/`, though `index check`, `render`, `orphans` and `coverage` all
do. The link is now plain code; the missing section is a real docs gap and is
not this PR's to close.

Gate on `840906d`: `compile`, `index`, `check --fail-on-unresolved
--fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` all exit 0.

bartekus:
Round 2, `106dd81`. Both findings accepted; both trace to round 1 rather than
to the original page, which is worth saying plainly.

**1. "stale shard" vs "drifted shard" — accepted.** The stronger version of
this: round 1 put `blocking-diagnostics` back into the class list, which made
"each stale shard" actively wrong, because the list again contains a class
that is not a stale shard. "Drifted" is also the word the implementation uses
(`drift_verdict`, the drift vector, the drift lines). Reverted to "each
drifted shard".

**2. The `3.3` section pin — accepted.** Checked before changing: it was the
only section pin anywhere under `website/docs/`. Every other reference in this
file is a bare id (spec 050, 086, 093, 094, 095, 097), and the tree uses no
section marks at all. Now reads `spec 098`; the paragraph already anchors the
contract in 098's acceptance rather than in a section number.

Net diff against `main` is now 7 insertions and 1 deletion: the intro sentence
is byte-identical to what it was before this PR, and the only replaced line is
the `blocking-diagnostics` bullet.

Gate on `106dd81`: `compile`, `index`, `check --fail-on-unresolved
--fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` all exit 0.

This is the second remediation round, which is the bound `/shepherd` sets. If
the next scan raises further prose preferences I will report them rather than
spend a third.

bartekus:
Round 3, `59cac8d`. One finding taken; three confirmation requests answered
with evidence, since each was answerable from the repository.

**Wording nit — accepted, fixed.** A bullet that calls itself "not printed in
prose" while being read in prose is confusing, and the intended contrast is
human-readable CLI output against the machine surfaces. It now reads "Not
emitted in the human-readable CLI output since spec 098".

**"Confirm 098 is not implemented by this PR" — confirmed, and the premise
behind it is not a thing here.** Spec 098 was built and merged in #215 (squash
`63ae87a`) and ratified in #217 (`485c11a`). On `main` it reads
`status: approved`, `implementation: complete`. This PR is documentation
catching up, so no spec file belongs in the diff and no waiver is needed.

For future scans: there is no `status: planned` in this project. `Status` in
`crates/spec-spine-types/src/frontmatter.rs` has exactly four variants, `draft`,
`approved`, `superseded` and `retired`, and `status: complete` does not exist
either. `implementation` is the separate axis carrying `complete`. `planned` is
a real word here but it is a *unit* flag from spec 076, not a lifecycle status.

**"Confirm the machine verdict collapses to one entry per shard" — confirmed
in source.** `IndexFreshnessReport::freshness()` in
`crates/spec-spine-core/src/index.rs` maps the blocking claims to `c.shard`,
collects them through a `BTreeSet<String>`, and only then formats
`blocking-diagnostics {file}`. The de-duplication is the set, so the collapse
is structural rather than incidental. Its doc comment states the intent: one
line per shard rather than one per diagnostic, so the count line means what it
meant before.

**"Confirm `spec-spine index diagnostics` is the exact subcommand name" —
confirmed.** `spec-spine index --help` lists `diagnostics  List the
diagnostics the committed index records (spec 050)`. The same page already
referenced `index diagnostics --json` before this PR.

Gate on `59cac8d`: `compile`, `index`, `check --fail-on-unresolved
--fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` all exit 0.

PR #221, opened 2026-09-16
chore(099): ratify (draft -> approved)

bartekus:
Ratify half of this repository's file-draft, build, ratify cadence, for spec 099 (#220).

`status: draft` becomes `status: approved`. Nothing else changes: `implementation` was already `complete` when #220 merged, and only the lifecycle key plus its two shards move.

**Why it is ratifiable now**

- #220 is merged (`4fab617`) and confirmed on disk.
- `spec-spine verify 099` passes on the merged default branch: 5 commands, exit 0, including the three shipped hook copies' parity check and the two body assertions.
- Gate on this branch, exit codes read unmasked: `check --fail-on-unresolved --fail-on-warn` 0, `lint --fail-on-warn` 0, `index coverage --fail-on-untraced` 0, `couple --base origin/main --head HEAD` 0 (1 path checked, no drift).

**What ratification does not settle**

Design note 06 §3.9's **H-6** (does the `Stop` hook refuse, or advise?) stays open. 099 §1.4 pins the advisory posture as it stands and a test holds it across all six exit paths, so H-6 remains a human decision rather than something an approved spec has quietly answered.

Corpus goes to **100 of 100 approved** (`registry list --json`, counted on this branch).


bartekus:
Answering the automated review's two points, since both are answerable from the repository's own documented cadence rather than from this 2-line diff.

**Same-day creation and approval is this repository's cadence, not a skipped step.** `AGENTS.md` "Working the backlog" opens with it: *"This repository files a spec as `draft`, builds it, then ratifies it in a separate PR, so the loop here differs from an adopter's ratify-then-build corpus in step 1 and step 6."* Step 6 makes the ratify PR the second of the two. Design note 05 §9.3 records the same thing from the other direction: `draft` plus `implementation: complete` compiles and lints clean here **by design**, because a lint refusing that pair would refuse this corpus's normal in-flight state. An adopter whose cadence is ratify-then-build would want the opposite, and that is noted there as a configurable lint rather than a default.

**The implementation did land in an earlier commit.** It is #220, squashed as `4fab617`, merged before this branch was cut. The reviewer's caveat ("it may have landed in an earlier commit") is the correct reading.

The gates are configured and were not bypassed. On the merged default branch, `spec-spine verify 099` passes its declared `## Verification` block, 5 commands, exit 0, which is the verb an orchestrator's verify stage runs after merge (spec 049). On this branch, read unpiped: `check --fail-on-unresolved --fail-on-warn` 0, `lint --fail-on-warn` 0, `index coverage --fail-on-untraced` 0, `couple --base origin/main --head HEAD` 0. CI's `self-governance` job re-ran all of it here and is green, as are the four determinism triples and the byte-identical assert.

One thing the review could not have seen and which is worth stating on the record: the fourteen tests spec 099 adds were run against the pre-099 hook bodies before the fix landed. Eight fail there, six pass as invariants the change must not move, and the three-copy parity test was separately shown to fail when only two of the three copies were updated. The acceptance can fail, which is the property that makes ratifying it mean anything.

PR #236, opened 2026-09-17
chore(105): ratify (draft -> approved, pending -> complete)

bartekus:
Ratification for spec **105**, filed and merged in #235.

## What this flips, and why it flips two axes

| Axis | Before | After |
|---|---|---|
| `status` | `draft` | `approved` |
| `implementation` | `pending` | `complete` |

The 100 to 104 series each ratified with a `status` flip alone, because the
`implementation` flip had already happened in that spec's build PR. Spec 105
has no build PR: it establishes no code, so the spec document was the whole
change and there is nowhere else for the second flip to live. Leaving it at
`pending` would have made 105 the only spec in a 106-spec corpus not marked
complete, describing work that is already in the tree.

## The state this records was already true

The substitution took effect when #235 merged, because `verify` resolves an
amendment from the corpus rather than from either lifecycle axis. On `bca8aef`,
with 105 still `draft` and `pending`:

```
$ spec-spine verify 098
verify: 098-a-blocking-claim-is-not-a-stale-shard
  acceptance amended by 105-an-amendment-carries-the-acceptance-it-replaces (spec 040); its block is the one that runs
...
verify: 098-a-blocking-claim-is-not-a-stale-shard: passed (60 command(s))
```

That block had been red since spec 101 merged, through five releases of the
corpus. This commit records on the axes what the tree already does.

## Scope

No file under `crates/` changes, no schema constant moves, no CLI surface is
added or removed. Three files: 105's `spec.md` and the two regenerated shards
its frontmatter produces.

## Testing

- `check --fail-on-unresolved --fail-on-warn`: both halves fresh, exit 0
- `lint --fail-on-warn`: 0 errors, 0 warnings, 0 info, across 106 specs
- `index coverage --fail-on-untraced`: 47/47, 22/22, 5/5 claimed, 0 unclaimed
- `couple`: 1 path checked, no drift
- `cargo test --workspace --locked`: 0 failures; `clippy -D warnings` and
  `fmt --all --check`: clean
- `spec-spine verify 105`: passed (60 commands)


bartekus:
Both findings are conditional on something the 4-line hunk cannot show. Answering them with the tree rather than with the diff.

## 1. `implementation: complete` without observable implementation artifacts

The first of the two interpretations is the right one: **spec 105 is purely declarative.** Its corrected acceptance block was authored and merged in #235; this PR moves the axes and nothing else.

Read through the CLI rather than off the shard:

```
$ spec-spine registry show 105 --json
establishes:        None
units:              0
dependsOn:          ['098-...', '101-...', '103-...']
amends:             ['098-...']
amendsVerification: ['098-...']
```

Spec 105 claims **no units at all**, so there is no code path that could be missing. That also settles the spec 041 concern directly: what `implementation: complete` removes is draft leniency on unresolved claims, turning a `W-001` into a blocking `I-004`. With zero units there is nothing for it to bite on, and the verb agrees on the flipped tree:

```
$ spec-spine check --fail-on-unresolved --fail-on-warn ; echo $?
0
```

On spec 049: `verify` is exactly the confirmation this finding asks for, and both blocks pass on this branch.

```
verify: 105-an-amendment-carries-the-acceptance-it-replaces: passed (60 command(s))
verify: 098-a-blocking-claim-is-not-a-stale-shard: passed (60 command(s))
```

One correction to the finding's framing: **CI does not run `verify`**, so "a false-complete claim would surface there" does not hold. `verify` is the one verb that executes rather than reads, and `AGENTS.md` keeps it outside the gate chain deliberately, because the chain runs against branches whose contents are in the general case a stranger's. That gap is not incidental here: it is the entire subject of spec 105. Spec 098's block was red from the day 101 merged, through five releases, with every gate green the whole time. The two runs above were done by hand, which is where that confirmation has to live, and §4 of the spec records the periodic sweep as the remedy rather than inventing a gate step.

## 2. No spec 098 cross-reference change

The body already names 098, in three places, which the `registry show` output above reads back: `dependsOn`, `amends`, and `amendsVerification`. Spec 103 §3.1 requires every `amendsVerification` entry to appear in `amends`, and that holds.

Spec 052 is misapplied here, though. It governs the **`C-001` refusal message**: it makes the coupling gate name the `extends` crossing when it refuses a changed path, so the author is not left with two shut doors. It places no requirement on a spec's frontmatter. The gate on this branch reports:

```
spec-spine couple: OK: 1 path(s) checked, no drift.
```

One term in the finding is worth not carrying forward: 105 does not **supersede** 098. `supersedes` and `amends` are distinct edges in this corpus, and the distinction is the point of spec 040: an amendment records the change **without touching the amended file**, which is why 098's `spec.md` is not in this diff and why its block still carries the superseded assertion form as a tripwire.

Nothing to change in the PR.

PR #237, opened 2026-09-17
docs(106): draft spec 106, replacing spec 050's stale acceptance

bartekus:
First entry in the acceptance audit spec 103 §4 named and left open.

## What is wrong

`spec-spine verify 050` has been red, and `verify` stops at the first failure, so only one of its three defects was visible. Running the block to the end finds three, and **none of them describes a defect in the code**:

| Cmd | Assertion | Actual | Amended by |
|-----|-----------|--------|------------|
| 5 | `"$(index check)" = "index is fresh"` | a second line, `unwitnessed claims: 87 (…)` | 057 |
| 7 | `len(index diagnostics --json) = 0` over a bare array | `{"items": [], "schemaVersion": "0.1.0"}` | 093 |
| 8 | `index check --json` `schemaVersion` `= "0.2.0"` | `0.4.0` | 088 |

Every rule spec 050 states still holds. Specs 057, 088 and 093 each changed an output 050's acceptance asserts on; none declared `amends_verification`, and none could have, since spec 103 built that key and shipped it in v0.20.0 after all three had merged. The gate chain does not run `verify` (spec 049), so each crossing went red at merge time with nothing to report it.

## What this does

Declares 050's acceptance replaced through `amends_verification` and carries the corrected block. **Spec 050's file is not edited.** Three greps in the new block assert that 050 still carries its superseded assertion forms, so they go red the moment anyone resolves this by editing 050 instead.

Command 8 is the one worth naming as a class. Spec 050 §3.6 rules that the envelope version MUST NOT move for a payload addition, and a literal pin cannot assert that: it also fails for every move the rule permits, which is what spec 088 did. The replacement compares `schemaVersion` across two verbs whose payloads differ.

That comparison is described in the spec as a **consistency check, not proof of §3.6**. Both verbs read one envelope constant, so a bump made for a payload reason would move them together and the line would still pass. The full rule is a statement about cause over time and is not decidable from one run of the tree. §3.3 states the boundary rather than leaving a reader to infer the line proves more than it does.

One assertion is **added**, on the `report.diagnostics` member. §3.6's rule has an antecedent that 050's own block never asserted, so the pin hung on nothing: drop the member entirely and both the original line and the consistency check still pass, while §3.6 has nothing left to govern. D-3 records the justification for adding to a block being repaired.

Both verb outputs are captured to files rather than compared through command substitution. `test "$(A)" = "$(B)"` passes when A and B both fail and both substitutions are empty. For the same reason the verdict-line assertion redirects to a file instead of piping into `head`, which exits 0 on empty input and would report the wrong defect (D-4).

## Fail-first

Only `registry show 106` is red at the parent commit. The corrected lines are not, and measurably so: this spec changes no code, so they exit 0 against the parent's binary. Each was instead measured against the condition it exists to catch and failed there; D-5 records the probes.

D-5's first draft claimed the corrected lines were fail-first and then described the *original* lines failing. Running them caught it. That is the fourth instance in this corpus of an acceptance claim written from the argument rather than measured, and the first where the unmeasured claim was about the fix rather than the defect; D-5 records it so the pattern is recognised rather than rediscovered.

## Verification

```
spec-spine verify 050   -> passed (20 commands)
spec-spine verify 106   -> passed (20 commands)
compile / index / lint --fail-on-warn / check --fail-on-unresolved --fail-on-warn / index coverage --fail-on-untraced   -> all 0
```

`verify 050` prints the substitution line spec 103 §3.4 requires, naming 106 as the holder.

## Scope left alone

Four more blocks (056, 059, 060, 071) are red for the same family of reason and each needs its own amendment: `amends_verification` replaces a target's whole block with the amender's single one, so a shared block would couple five unrelated acceptances. Recurring sweep coverage, and teaching `amends_verification` in the spec template, are both named in §4 as separate work.

`status: draft` is deliberate. Approval is a human flip.


bartekus:
Thanks: two of the three land, and both are now fixed in c0a8278. I verified each against the source before acting rather than taking the finding as given.

**Bug: `cargo test` filter can vacuously pass. Confirmed, fixed.**

Measured directly:

```
$ cargo test -p spec-spine-core --test verify --locked zzz_no_such_test_
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 31 filtered out
$ echo $?
0
```

`spec103_` does match 5 tests today, so the line asserts something right now, but your point is about the rename, and it is right. This is the vacuous-pass family this corpus has already paid for once. The run is now captured and its summary asserted to name a non-zero pass count, so the line goes red when the filter matches nothing as well as when a match fails. Verified both directions: guard exits 1 against `zzz_no_such_test_`, 0 against `spec103_`. Recorded as D-7.

One correction to the framing: the suggested `--exact` on a fully-qualified name would close it for one test and lose the other four. Asserting the count keeps the filter's breadth and still refuses an empty match.

**Logic: key-set comparison is itself temporally fragile. Confirmed, fixed.**

This is the better finding, because it lands on the spec's own argument rather than beside it. §1.2 and D-2 reject a literal pin for being a claim about the calendar, and `sorted(a['report']) != sorted(b['report'])` is the same shape on a different axis: it goes red if `compile`'s report ever gains the members `index check` carries, which no rule here forbids. So the assertion changed rather than the argument.

I did not take the suggested form as-is. Asserting only that `diagnostics` is absent from `compile`'s report is a single negative claim, and a negative about another verb's payload is weak ground. The replacement is two positive claims:

```python
assert a['verb'] != b['verb'], (a['verb'], b['verb'])
assert 'diagnostics' in a['report'] and 'diagnostics' not in b['report'], ...
```

The verb token is part of the envelope under spec 037, so `index.check` against `compile.spec` is a statement about what the two verbs *are*, not about what they happened to contain when this was written. That is the invariant the §3.3 argument actually needs. §3.3 now states the rule and forbids the key-set form explicitly; D-6 records the exchange.

**Minor: shell-expanded `TMPDIR` inside a Python string literal. Accurate, not changed.**

The mechanism is real: a single quote in `$TMPDIR` would produce a Python `SyntaxError`. Not changing it, for one reason: it fails loudly rather than silently. Every other finding here is about a line that can pass while asserting nothing, and that is the direction this corpus refuses. A misleading error message on a pathological `TMPDIR` costs a reader one minute; a green line that asserts nothing costs a release.

It is also the idiom spec 105's block uses. Diverging in one spec would leave two spellings of the same fixture pattern and no rule saying which is current. If it is worth fixing it is worth fixing in both, and that is an amendment of 105, not a repair of this block.

**Non-findings section:** accurate. The redirect-then-`head` and two-file comparison readings are both correct, including why the second one matters (`test "$(A)" = "$(B)"` passes when both verbs fail and both substitutions are empty).

Gate re-run to completion on the new head: `compile`, `index`, `lint --fail-on-warn`, `check --fail-on-unresolved --fail-on-warn`, `index coverage --fail-on-untraced` all 0. `verify 050` and `verify 106` pass at 21 commands each.

bartekus:
Second pass: one finding actionable, two declined with evidence. Fixed in a096d53.

**1. `compile --spec 024` is a brittle anchor. Half actionable, taken.**

The deletion arm is fair and the fix is free: the counter-verb now anchors on **spec 000**, the tier-1 bootstrap. Any id can be deleted, so the anchor is a dependency whichever one is chosen, but 000's `unamendable` anchors are non-overridable and its removal would end the corpus rather than break this line. Payload shape is identical (`['passed','specId','specPath','violations']`, `verb=compile.spec`), so nothing else moves. 024 was the first draft's arbitrary pick, and you were right to flag it as unexplained; D-8 now names the choice so a future maintainer can substitute correctly.

The payload-growth arm lands on the spec's own text, which is the more useful half. D-6 said both arms describe "what the verbs are, not what they happened to contain". That is an overclaim: `'diagnostics' not in b['report']` is a claim about what `compile` happens to contain. In a spec whose entire subject is assertions that overclaim, the sentence is corrected rather than defended, and D-8 records what the arm actually rests on.

The arm stays, and the reason is the failure mode rather than the contingency. A literal version pin goes red on a routine, expected event: spec 088 performed one, which is how this whole audit started. This goes red only if two verbs' payloads converge on a named member, which no rule forbids but nothing plans either, and it goes red loudly and sends the reader to D-8. Contingency is not the defect; passing while asserting less than you appear to is.

**2. `$TMPDIR` interpolation. Declined, as in pass 1.**

This is a verbatim restatement of the previous pass's minor finding, answered [above](https://github.com/statecrafting/spec-spine/pull/237#issuecomment-first). The mechanism is real and the reason stands: it fails loudly with a `SyntaxError`, never silently, and it is the idiom spec 105's block uses. Diverging in one spec leaves two spellings of the same fixture pattern and no rule saying which is current. If it is worth fixing it is worth fixing in both, which is an amendment of 105.

**3. Non-unique scratch directory. Declined; the citation does not hold.**

The finding attributes the concern to "the spec's own narrative in §1.3". §1.3 is titled "The crossings were silent, and nothing reruns a merged block" and is about specs 057, 088 and 093 changing outputs without declaring `amends_verification`. It says nothing about scratch directories or silent failures from leftover data. Please do not attach a finding to a section that does not support it: a reader who follows the pointer finds an unrelated argument, and the misattribution costs more to run down than the finding is worth.

On the substance, the static path is the corpus convention, not an oversight: 17 verification blocks use `${TMPDIR:-/tmp}/ssNNN` with a per-spec ordinal, which is what makes them collision-free between specs. `verify` executes one command at a time within one block, and the ordinal scopes the directory across blocks. Two concurrent runs of the *same* spec would race, which is true of every one of the 17 and is a property of the corpus convention, not of this spec.

**Compliance section:** accurate on all five points.

Gate re-run to completion: `compile`, `index`, `lint --fail-on-warn`, `check --fail-on-unresolved --fail-on-warn`, `index coverage --fail-on-untraced` all 0. `verify 050` and `verify 106` pass at 21 commands each.

bartekus:
Third pass. Nothing actionable; the first finding is testable and the test refutes its conditional. Both remediation rounds are spent, and none of these is a defect, so this is where I stop.

**1. `amends_verification` + `status: draft`. The conditional is false, measured.**

The concern is that `cmd_verify` might gate the lookup on a non-draft status, in which case `verify 050` "continues to fail after this PR merges". It does not, and this is directly observable on the branch, where 106 is `draft` / `pending` right now:

```
$ spec-spine registry show 106 --json   ->  status: draft | implementation: pending
$ spec-spine verify 050
verify: 050-index-diagnostics-reach-a-gate
  acceptance amended by 106-... (spec 040); its block is the one that runs
...passed (21 command(s))
$ spec-spine verify 050 --plan | wc -l   ->  21     # 050's own block is 8 commands
```

A draft amender already holds the acceptance, which is why the plan is 21 lines and not 8.

It is also settled in the mechanism's own spec rather than left open. Spec 103 §3.2 excludes exactly two statuses and says why: "A `superseded` or `retired` amender MUST be skipped: its acceptance is no longer the corpus's." Draft is not in that set, and an amender that had to be ratified before its block took effect would leave the target red during review, which is the window where the block most needs to run. Since 103 answers it normatively and the branch demonstrates it, there is no gap in §4 to name: §4 lists what this spec declines to do, not restatements of its dependencies' rules.

**2. Scratch directory not process-isolated. Third statement of this, declined.**

Same finding as pass 2 §3 and pass 1's neighbourhood, now with the qualifier "likely accepted practice", which is correct: 17 blocks in this corpus use `${TMPDIR:-/tmp}/ssNNN` with a per-spec ordinal.

One correction: the finding cites spec 105's path as `ss105`. It is `ss098`, because the directory is named for the spec whose acceptance the block holds, not for the holder. That detail matters here, since it is the convention that makes these paths collision-free between specs, and a reviewer inferring the naming rule backwards would propose the wrong fix.

**3. Pipeline masks `registry show` exit status. Accurate, and correctly self-classified.**

The analysis is right on every point, including that fail-first is preserved and that the direction is false-diagnostic rather than false-pass. That is the distinction that decides it: D-4 exists because a masked failure can pass while asserting nothing, and this shape cannot. A Python traceback instead of a CLI diagnostic costs a reader one command to reproduce; it never costs a release. Left as it is, deliberately.

**Non-findings section:** accurate, including the read of the D-7 guard regex and the §3.2 redirect split.

Merging on nine green checks at `a096d53`.

PR #239, opened 2026-09-17
docs(107): draft spec 107, replacing spec 056's stale acceptance

bartekus:
Files spec 107, `status: draft`. First of the four follow-ons spec 106 §4 named.

## What is red, and why it is not a defect

`spec-spine verify 056` has failed at command 5 since spec 088 merged:

```
assert v["schemaVersion"]=="0.3.0"   # the envelope answers 0.4.0
```

088 moved `VERDICT_SCHEMA_VERSION` from `0.3.0` to `0.4.0` when `delta` was added: the additive MINOR case that constant's own documentation names, and the case spec 056 §3.4 itself says governs future bumps. The pin was always stricter than the rule it stood for.

Running 056's block one line at a time to the end, command 5 is the only red one; commands 6 to 8 are green.

## The pin asserted neither half of §3.4

- **Too strong for the version.** 056 §3.4 records the `0.2.0 -> 0.3.0` bump in the past tense and states the rule for future ones. A pin refuses what that rule permits. Spec 105 D-2's test: an assertion that must eventually fail for a legitimate reason is not an assertion.
- **Too weak for the verb.** §3.4's requirement is a `verb` token *distinguishing* `--spec` from `compile --check`. 056's block never ran `compile --check --json`. Give both verbs the token `compile.spec` and the block stays green while the requirement breaks.

## What this PR does

No code. 107 declares `amends` + `amends_verification` on 056 and carries the replacement block, per spec 103 §3.1. Spec 056's file is not edited (spec 040 §3.1), and the replacement asserts that it still carries the superseded pin, so the line goes red if anyone ever resolves this by editing 056 instead.

The replacement:
- compares the envelope version **across the pair §3.4 names** (`compile.spec` vs `compile.check`) instead of pinning a literal, and reads both verb tokens rather than one;
- redirects each document to a file so the verb's own exit status is the line's status (a pipeline reports `python3`'s);
- adds two payload assertions: `--spec`'s report names the resolved full id, path and empty `violations` (which is where 056 §3.1's short-id resolution actually shows: the exit-code line passes whatever `024` resolved to), and `compile --check`'s report carries no `specId`.

## Failability evidence

This spec changes no code, so the corrected lines pass at the parent commit. That is spec 106 D-5's finding at a second site, and D-5 here records what each line was measured against instead:

| Condition | Line |
|---|---|
| the two envelope versions diverge | fails |
| `--spec` answers the token `compile.check` | fails |
| `--spec` resolves `024` to another spec's id | fails |
| `compile --check`'s report gains a `specId` | fails |

`registry show 107` is the one line fail-first at the parent in the ordinary sense (not-found exit 1).

## Gate

```
spec-spine verify 056   -> passed (19 command(s))   # runs 107's block
spec-spine verify 107   -> passed (19 command(s))
```


bartekus:
Round 1 (8ccbd0f). One finding fixed, one declined with evidence.

**Fixed: the piped `registry show` line.** The finding is correct and it lands on this spec's own argument, which is the kind worth spending a round on. D-4 states the rule and the first draft broke it eleven lines later, in the one place where the hazard is not hypothetical: at the parent commit `registry show 107` exits 1 and prints nothing, so the pipeline reports a JSON decode error for what is a not-found. That is exactly the misattribution D-4 exists to prevent, met by the first reviewer who runs the block.

The line is now a redirect to `${TMPDIR:-/tmp}/ss107-show.json` and a read from the file. The path is named for **107**, not 056: it is this spec's own mechanism assertion, not part of 056's acceptance, and D-6 draws exactly that distinction. D-4 is extended to record the fix and why the rule reaches every line, and §4 now names the same form in specs 105's and 106's blocks, which are green (on a merged tree the verb succeeds) and are another spec's acceptance to change.

**Declined: `assert a['schemaVersion'], a` raising `KeyError`.** An absent member and an empty one are different defects, and `KeyError: 'schemaVersion'` names the first precisely. `a.get(...)` would replace a message that identifies the missing member with one that does not, in exchange for printing the document. The line goes red either way, which the review grants. The form is also what spec 106's block uses at the same assertion, and diverging here would leave the family inconsistent for no gain.

Evidence after the fix:

```
spec-spine verify 056   -> passed (21 command(s))   # runs 107's block
spec-spine verify 107   -> passed (21 command(s))
```

bartekus:
Round 2. The redirect fix is confirmed; the remaining note is declined a second time, now with the evidence that answers its sharper form.

**Declined: `assert a['schemaVersion'], a` raising `KeyError`.** The round-2 argument is better than the round-1 one: the concern is not the exception type but that the diagnostic argument `a` is never printed on the missing-key path. That is true of the expression in isolation and not true of this block, because of the fix made in round 1.

§3.2 requires each document to be **captured to a file** rather than piped, and `verify` stops at the first failing command, so the `rm -rf` two lines below never runs. When that assertion fails, both envelopes are sitting on disk at `${TMPDIR:-/tmp}/ss056/spec.json` and `check.json`, complete and pretty-printed, which is strictly more than `assert a.get(...), a` would print. The argument dump is a convenience here, not the diagnostic of record.

On the exception itself: `KeyError: 'schemaVersion'` names the absent member exactly, and an absent member and an empty one are different defects. `.get()` collapses them into one message that identifies neither.

The form is also what spec 106's block uses at the same assertion, and specs 108 and 109 carry it for the same reason; diverging at one of the four would leave the family inconsistent for a diagnostic that is already on disk.

All nine checks green at 8ccbd0f. Merging.

PR #240, opened 2026-09-17
docs(108): draft spec 108, replacing spec 059's stale acceptance

bartekus:
Files spec 108, `status: draft`. Second of the four follow-ons spec 106 §4 named.

## What is red, and why it is not a defect

`spec-spine verify 059` has failed at command 4 since spec 093 merged:

```
assert set(o) == {"orphaned", "inFlight"}
```

`index orphans --json` answers `{inFlight, orphaned, schemaVersion}`. 093 routed every read document through one emitter that sorts keys and stamps a `READ_SCHEMA_VERSION`, and named this verb in its own D-8. Running 059's block one line at a time to the end, command 4 is the only red one; the other six are green.

## The equality asserted a closed document 059 never stated

059 §3.1 says the `--json` form is "an object with the two arrays as named members" and "always emits both arrays". Both sentences are about **presence**. A key-set equality asserts that nothing else may ever appear, which contradicts the schema policy it sits among (MINOR is the additive case; the embedded `specRecord` sets no `additionalProperties` deliberately) and which 093 then legitimately violated.

It is also weaker than it looks: `set(o)` says nothing about the members' **types**. Replace both arrays with strings and the line still passes, while the shape 059 exists to produce is gone. A consumer's breakage is a wrong type, not an extra key.

## The block was also silent about 059's own two decisions

059 §3.1 carries two dated decisions the block never asserted:

- `--json` emits **both** arrays even when empty, so a consumer never distinguishes absent from empty. Command 4 ran against this repository, where both groups are populated, so the case the decision is about was never exercised.
- the prose form prints **nothing** on a corpus with no orphans ("two headers and two `(none)` lines would be noise"). Nothing in 059's block read the prose form at all.

The block already builds an empty-universe fixture for §3.2/§3.3, and that fixture is exactly the corpus both decisions describe. Two lines, no new machinery.

## What this PR does

No code. 108 declares `amends` + `amends_verification` on 059 and carries the replacement block, per spec 103 §3.1. Spec 059's file is not edited (spec 040 §3.1), and the replacement asserts it still carries the superseded equality, so that line goes red if anyone resolves this by editing 059 instead.

§3.6 records that spec 093 owed 059 **no** `amends` edge: 093 amended 010 and 060 because each states a document's shape as a contract it broke; 059 states presence, and 093 preserved presence. An over-strict assertion is not a rule anyone amended. The opposite reading would make every additive change an amendment of every spec whose block happened to be written tightly.

## Failability evidence

This spec changes no code, so the corrected lines pass at the parent. D-5 records what each was measured against instead:

| Condition | Line |
|---|---|
| pre-059 output, a bare array of ids | fails (`TypeError`) |
| pre-093 output, no `schemaVersion` | fails |
| keys in struct order rather than sorted | fails |
| a group omitted from the document | fails |
| an empty group reported as non-empty on the fixture | fails |
| the prose form run on a corpus that has orphans | fails |

D-3 records one thing deliberately **not** added: 059 §3.1 requires each group to stay id-sorted and nothing checks it, but this repository's groups hold one or two entries and the fixture's hold none, so the assertion would pass whatever the code did. Adding a line that cannot fail is worse than leaving the gap visible; §4 names it.

## Gate

```
spec-spine verify 059   -> passed (26 command(s))   # runs 108's block
spec-spine verify 108   -> passed (26 command(s))
```


bartekus:
Round 1 (7c9670a, with 107's merge cfaad4e merged in). One finding fixed, one declined with evidence.

**Fixed: spec 107 referenced as merged but absent from the base.** The finding is correct, and the first of its two branches is the true one: 107 was a concurrent PR (#239), now merged, and this branch now carries it. The review's real point survives that, which is that the ordering was undeclared. It is now declared: `107-a-version-pin-is-not-a-contract` is in `depends_on`, with a frontmatter comment and a new D-7 recording that the dependency is on 107's **record**, not on its behaviour. Nothing here executes, reads or relies on anything 107 does, and the two blocks share no state; what this spec borrows is §1.2's list of corrected sites, §3.2's redirect precedent, and §4's position in the series. The cost is that `registry plan` marks 108 blocked until 107 is complete, which is the truth of it.

**Declined: the `cargo test` redirect misdirecting a reader.** The described misdirection cannot occur, because the redirect does not swallow the status. `cargo test ... > f 2>&1` leaves **cargo's own exit status** as that line's status, and `verify` stops at the first failing command. A panic or a cargo error takes the line down where it happens and `verify` names the `cargo test` line; the `grep` below only runs when cargo exited 0. The case the `grep` exists for is the one cargo reports as success: `0 passed; 31 filtered out`, status 0, measured by spec 106 D-7. The block therefore already has the separate-status treatment §3.2 asks for, and the captured file is on disk either way, because the `rm -f` below it never runs after a failure.

Evidence after the fix:

```
spec-spine verify 059   -> passed (26 command(s))   # runs 108's block
spec-spine verify 108   -> passed (26 command(s))
```

bartekus:
Round 2 (ed11d93). One finding fixed, one turned into a spec change, one declined with evidence.

**Fixed: D-6 and D-7 out of order.** Correct, and it is the kind of defect that matters in a document whose subject is legibility: both entries have live cross-references (the Verification block cites D-6 twice, the frontmatter comment cites D-7), and a reader following either scrolled past the other. Caused by inserting D-7 ahead of D-6 in the patch that added it. The log now reads D-1 through D-8 in order.

**Acted on: the `index coverage … | grep -q` line in tension with §3.2.** The observation is fair and the tension was in the spec, not the block. §3.2's first draft said the rule reached "**every** line", and that is wrong: the boundary is which status the line should report.

- Where a verb must **succeed**, its status is the line's answer, and a pipeline throws it away. That is the rule, and it now reads "every line whose verb must succeed".
- Where a verb must **refuse** — `index coverage --fail-on-untraced` on an empty universe, where exit 1 *is* what 059 §3.2 requires — the non-zero status is the expected outcome and the question is what the refusal said. There `grep`'s status is the right one to surface, and a redirect would make the line fail on a correct refusal.

§3.2 now states both halves and new D-8 records why, so the inherited line reads as a decision rather than an oversight.

**Declined: `assert d['schemaVersion'], d` as truthiness rather than a type check.** `READ_SCHEMA_VERSION` is `pub const READ_SCHEMA_VERSION: &str` in `crates/spec-spine-types/src/version.rs`, and every read document is stamped from it through the one emitter spec 093 §3.3 routes them all through. The member's type is therefore a Rust compile-time fact; a non-string could only arrive by changing that const's type, which the compiler would force through every call site before any block could run. An `isinstance` check in a shell line cannot be the guard for a Rust type, and §3.3's own words are "a non-empty `schemaVersion`", which is exactly what truthiness asserts.

Evidence after the fixes:

```
spec-spine verify 059   -> passed (26 command(s))   # runs 108's block
spec-spine verify 108   -> passed (26 command(s))
```

bartekus:
Round 2 review. All three notes are declined, two on evidence and one as a style preference on an inherited line.

**1. Truthiness rather than a type check.** Standing reply above; the review grants it is low severity, and nothing new is named. `READ_SCHEMA_VERSION` is `pub const READ_SCHEMA_VERSION: &str`, so the member's type is a Rust compile-time fact rather than something a shell line can guard.

**2. The empty-fixture assertion omitting `schemaVersion`.** This one is new and worth answering properly. The omission is not a gap, because there is no empty-corpus path to regress separately. `crates/spec-spine-cli/src/cmd_index.rs:115` is the whole of it:

```rust
if *json {
    out!("{}", read_document(&report, Versioning::Stamp)?);
}
```

One unconditional call, no branch on whether the groups are populated, and `orphans_json` in `core/src/lib.rs:479` does the same for the facade. The version member is stamped by the same line of code in both invocations, so "a regression that drops `schemaVersion` from the empty-corpus path" is not a state the code can be in: it would have to drop it from both, which the live-corpus assertion catches. What differs between the two invocations is the *contents* of the two arrays, and that is what the empty-fixture line asserts, because that is the only thing that could differ.

Adding the assertion would cost nothing and witness nothing, which is the standard D-3 applies to the id-ordering line it declines to add: a line that cannot fail is worse than no line, because a reader counts it as cover.

**3. `; test $? -eq 1` wanting an inline annotation.** The line already carries a two-line comment directly above it saying what it expects and why ("with the flag it refuses. An assertion over an empty set is vacuously true..."). It is inherited from 059's block unchanged, and re-annotating another spec's line to a different house style is churn inside a diff whose whole claim is that it changed one assertion and added two.

All nine checks green at ed11d93. Merging.

PR #238, opened 2026-09-17
chore(106): ratify (draft -> approved, pending -> complete)

bartekus:
Flips spec 106 from `draft` to `approved` and `pending` to `complete`.

106 is a no-build spec: it replaces spec 050's stale `## Verification` block through `amends_verification` and changes no code, so both lifecycle axes move together (the shape spec 105's ratification used).

Evidence on this branch:

```
spec-spine verify 050   -> 0   (runs 106's block, per spec 103 3.2)
spec-spine verify 106   -> 0
spec-spine check --fail-on-unresolved --fail-on-warn -> 0
spec-spine lint --fail-on-warn -> 0
spec-spine index coverage --fail-on-untraced -> 0
spec-spine couple --base origin/main --head HEAD -> 0 (1 path checked, no drift)
cargo test --workspace --locked -> all green
```

Only 106's own registry and index shards move.

PR #241, opened 2026-09-17
chore(107): ratify (draft -> approved, pending -> complete)

bartekus:
Flips spec 107 from `draft` to `approved` and `pending` to `complete`.

107 is a no-build spec: it replaces spec 056's stale `## Verification` block through `amends_verification` and changes no code, so both lifecycle axes move together. Same shape as specs 105 (#236) and 106 (#238).

Evidence on this branch:

```
spec-spine verify 056   -> passed (21 command(s))   # runs 107's block, per spec 103 3.2
spec-spine verify 107   -> passed (21 command(s))
```

Only 107's own registry and index shards move.


bartekus:
The finding is answered; two of its premises need correcting, and one of them is the thing this series exists to document.

**The artifact exists at base, and it is the block itself.** 107 is a no-build spec: its artifact is its `## Verification` block, which becomes spec 056's acceptance through `amends_verification` (spec 103 §3.2). It merged in cfaad4e, which is this branch's base. The evidence that it works is the verb, not the diff:

```
$ spec-spine verify 056   # runs 107's block, not 056's
verify: 056-compile-one-spec: passed (21 command(s))
$ spec-spine verify 107
verify: 107-a-version-pin-is-not-a-contract: passed (21 command(s))
```

Before 107, that first command was `FAILED at command 5`. Specs 105 (#236) and 106 (#238) moved both axes together for the same reason.

**"The CI verifier will reject it" is not so, and its being not so is spec 107 §1.3.** Nothing in CI runs `spec-spine verify`. It is the one verb that executes what the corpus declares, so AGENTS.md keeps it outside the gate chain deliberately: the chain runs against PR branches whose contents are, in the general case, a stranger's. That is exactly why spec 088 could move `VERDICT_SCHEMA_VERSION` and leave spec 056's block red for five releases with every gate green, which is the crossing 107 repairs. A block can be red for months and CI never says a word.

**Spec 058 is not the pattern here.** 058 is "The defects heading has one spelling" (the directory name predates its subject). The mechanism is spec 103's `amends_verification`, shipped in v0.20.0.

PR #242, opened 2026-09-17
docs(109): draft spec 109, replacing spec 060's stale acceptance

bartekus:
Files spec 109, `status: draft`. Third of the four follow-ons spec 106 §4 named.

## What is red

`spec-spine verify 060` has failed at command 8 since spec 093 merged:

```
assert json.load(sys.stdin)=={"id":"001-alpha","title":"First thing"}
```

The verb answers `{"next": {…}, "schemaVersion": "0.1.0"}`. Running 060's block one line at a time to the end, command 8 is the only red one.

## This one is different from 107 and 108, and §1.2 says so

107 and 108 correct assertions that were **stricter than their rules**. This is not that. 060 §3.2 states the shape as a contract ("the single spec object rather than an array") and the block asserted exactly that, correctly. Spec 093 then **changed the contract**, deliberately: `plan --next`'s empty case emitted the bare literal `null`, which cannot carry a version member, so the pick moved under a nullable `next`. 093 §3.5 carries the replacement text and 093's frontmatter carries the `amends` edge naming 060.

So 060's rule was amended by the book. What 093 could not reach was 060's acceptance, because spec 040 forbids editing the amended file and the block lives inside it — spec 105 §1.2's finding at a second site. Spec 103 built `amends_verification` after 093 shipped.

D-2 records this explicitly, so a reader meeting three of these in a row does not take the family to mean acceptance blocks should assert less.

## The block never asserted 060's strongest requirement

060 §3.2: "`--next` MUST be a projection of `plan` and MUST NOT reimplement selection. The first element of the topological order is already the defined pick."

The fixture has **one** ready spec. Against a one-element ready set, first, last, greatest and any reimplementation return the same object. The rule 060 argues hardest for is the one its acceptance cannot see.

The fixture gains `003-gamma` (approved, pending, no deps), so `ready` holds two. The block then asserts `next == ready[0]` of the same corpus, and the inherited `ready[0]` line stops being a tautology and becomes a claim about order. Measured: a `--next` answering `003-gamma` fails both lines, and with one ready spec it fails neither.

## What this PR does

No code. 109 declares `amends` + `amends_verification` on 060 and carries the replacement block, per spec 103 §3.1. Spec 060's file is not edited (spec 040 §3.1), and the replacement asserts it still carries the superseded equality.

Also carried across unchanged: 060's own dated decision of 2026-09-08 that the shape assertions run against a scratch corpus because "the live corpus is an input, not a contract". §3.4 changes what that corpus contains, never where the assertions point. D-4 records the tempting cheap version that was **not** taken — asserting `plan --next`'s text against this repository because its ready set happens to be empty today, which is that exact defect one line below where the decision entry warning about it sits. The empty case is guarded by 093 §3.8 in `crates/spec-spine-cli/tests/cli.rs`, with a fixture built for it.

## Failability evidence

| Condition | Line |
|---|---|
| pre-093 output, the bare pick object | fails (`KeyError: 'next'`) |
| `next` holding a one-element array | fails |
| the pick answering `003-gamma` | fails, and so does the projection line |
| `ready[0]` and `next` disagreeing | fails |
| keys in struct order rather than sorted | fails |
| no `schemaVersion` member | fails |

## Carried over from the reviews of #239 and #240

- The `registry show` read is redirected, not piped (107 D-4).
- The two prose assertions are captured once to a file rather than piped twice, and §3.2 states the boundary 108 D-8 drew: the redirect rule reaches every line whose verb must **succeed**, and a line whose verb must refuse keeps its pipeline. This block has no such line.
- `107` and `108` are declared in `depends_on` (D-7), because this spec makes past-tense claims about both.

## Gate

```
spec-spine verify 060   -> passed (22 command(s))   # runs 109's block
spec-spine verify 109   -> passed (22 command(s))
```


bartekus:
Round 1 (6fa995a). No finding was blocking, and the review's closing suggestion is taken: all three assumptions are now in the decision log as D-8, rather than only in this thread. Two of the three framings need correcting, and the corrections are what D-8 records.

**`ready[0]` and the ordinal tie-break.** Real, and the load-bearing part is right: the two ready specs are independent, so `001-alpha` first is settled by the tie-break spec 053 governs. But the consequence is the opposite of "vacuous without a visible failure". `p['ready'][0]` is compared against the **whole object**, so a reordered plan reports `{'id': '003-gamma', …}` and the line goes red on the spot. The projection line beside it would keep passing, and correctly so: what it asserts is that `--next` follows `ready` wherever `ready` points, which is 060 §3.2's actual rule and is independent of which spec is first.

**`not schedulable` at a zero count.** Also real, and also loud rather than silent: if the CLI suppressed the section at zero, `grep -q` fails and `verify` stops there naming that line. "Silently red" would need the line to pass while the property was gone, which is the failure direction this whole series is about and is not this one. Measured again on the extended fixture: `3 specs: 2 ready, 1 blocked, 0 not schedulable`.

**The Rust test-harness output format.** Accurate and outside this project's control. It is the form spec 106 D-7 introduced, and the alternative is the bare filtered run that exits 0 on zero matches, which is the vacuous pass that entry exists to close. A dependency on an external format that fails closed beats an assertion that cannot fail.

D-8 states all three with the failure direction for each, so the next reader finds them where the spec is rather than where the review was.

Evidence after the addition:

```
spec-spine verify 060   -> passed (22 command(s))   # runs 109's block
spec-spine verify 109   -> passed (22 command(s))
```

bartekus:
Round 2 (0fa83b2). The finding's conclusion is wrong and its target is right, so the fix is made rather than the objection dismissed.

**The conclusion.** `NameError` is not what command 8 was failing on. Spec 060's file carries the imports, at `specs/060-plan-answers-the-whole-question/spec.md:238`:

```
target/release/spec-spine --repo "${TMPDIR:-/tmp}/ss060" registry plan --next --json | python3 -c 'import json,sys; assert json.load(sys.stdin)=={"id":"001-alpha","title":"First thing"}'
```

and the measured failure at `e4d7d28` is the comparison, not the import:

```
AssertionError
[verify] exit 1
verify: 060-plan-answers-the-whole-question: FAILED at command 8 (exit 1)
```

So §1.1's account of the redness stands, and so does row 8 of the table.

**The target.** The review was reading my §1.1, not 060's file, and §1.1 had elided `import json,sys;` and the `--repo` argument for width. A quotation that changes what a command does is not a quotation, and the inference the review drew from it was a reasonable one to draw. In a spec whose subject is assertions that say less than they appear to, an elision that makes an assertion appear to say *more* than it does is the same error pointing the other way.

§1.1 now reproduces the line verbatim, so the next reader can run it, and D-9 records the exchange.

Evidence after the fix:

```
spec-spine verify 060   -> passed (22 command(s))   # runs 109's block
spec-spine verify 109   -> passed (22 command(s))
```

bartekus:
Round 3. Two of the three notes close themselves ("confirmed-safe, no change needed" and "accept as-is"); the third is testable and the test refutes it. No further change.

**1. "Draft status may suppress `amends_verification`, leaving 060 red after merge."** Refuted by running it. 109 is `status: draft` on this branch right now, and:

```
$ spec-spine verify 060
verify: 060-plan-answers-the-whole-question: passed (22 command(s))
```

060's own block is 11 commands; 22 is 109's. The substitution is already happening at draft.

It is also what spec 103 §3.2 says: a `superseded` or `retired` amender must be skipped, "because its acceptance is no longer the corpus's". `draft` is neither, and excluding it would break the build-then-ratify order this repository runs on (AGENTS.md step 6) — a spec is built and merged as `draft`, so an amender that only took effect at `approved` could never be verified before ratification. Specs 105, 107 and 108 each merged as draft and each was verified that way before its ratify PR. The same claim was raised on #237 and refuted the same way.

**2. Projection assertion.** The review works through both directions and concludes the assertion is tight, recording it as confirmed-safe rather than a finding. Agreed, and D-8 states the same boundary from the other side: the `ready[0]` value comparison catches a reordering, the projection line catches a `--next` that stops following `ready`.

**3. `spec103_` filter — confirmation requested before ratification.** Confirmed, five of them in `crates/spec-spine-core/tests/verify.rs`:

```
307: spec103_verify_runs_the_replacement_block
338: spec103_resolution_follows_the_chain
368: spec103_a_withdrawn_amender_does_not_hold_the_acceptance
395: spec103_a_spec_with_no_amender_runs_its_own_block
408: spec103_a_cycle_does_not_hang_verify
```

and the guard measures them: the run's summary reads `test result: ok. 5 passed`, which the `[1-9][0-9]*` pattern requires and which a filter matching nothing would not produce.

All nine checks green at 0fa83b2. Merging.

PR #243, opened 2026-09-17
chore(108): ratify (draft -> approved, pending -> complete)

bartekus:
Flips spec 108 from `draft` to `approved` and `pending` to `complete`.

108 is a no-build spec: it replaces spec 059's stale `## Verification` block through `amends_verification` (spec 103 §3.2) and changes no code, so both lifecycle axes move together. Same shape as specs 105 (#236), 106 (#238) and 107 (#241).

Nothing in CI runs `spec-spine verify`: AGENTS.md keeps it outside the gate chain deliberately, which is the crossing this whole series repairs. The evidence is the verb, run here:

```
spec-spine verify 059   -> passed (26 command(s))   # runs 108's block, per spec 103 3.2
spec-spine verify 108   -> passed (26 command(s))
```

Before 108, `verify 059` was `FAILED at command 4`. Only 108's own registry and index shards move.


bartekus:
Both notes are answered; the first rests on a premise about this repository that does not hold, and the citations need correcting.

**There is no third case missing: 108 is a no-build spec.** The review offers two cases and neither is this one. 108 ships no "enforcing logic", now or earlier. Its artifact is its `## Verification` block, which became spec 059's acceptance the moment it merged (spec 103 §3.2, `amends_verification`). The thing that was broken was an **assertion**, not code, and the repair is an assertion. `implementation: complete` on such a spec means its artifact exists and works, and the verb is the evidence:

```
$ spec-spine verify 059   # runs 108's block, not 059's
verify: 059-read-verbs-on-a-code-free-corpus: passed (26 command(s))
$ spec-spine verify 108
verify: 108-an-exact-key-set-refuses-what-the-rule-allows: passed (26 command(s))
```

Before 108 that first command was `FAILED at command 4`. Specs 105 (#236), 106 (#238) and 107 (#241) all ratified this way, and the coupling gate agrees: `couple --base origin/main --head HEAD` exits 0 with 1 path checked and no drift, because the one changed source path is 108's own `spec.md`, owned by 108.

**The acceptance block is present and unchanged.** It landed in 7ce016c (#240), which is this branch's base. A ratify PR is four frontmatter lines by design; the body it ratifies is deliberately not in the diff.

**Citations.** Spec 041 is "`implementation: complete` defeats draft leniency" — it makes a complete spec's claims *strictly* checked, which is the reason to run `verify` before flipping, and that was run. There is no "spec-041 (completion-held-to-claims)". Spec 042 is `attest --spec <id>`, a signed record scoped to one spec; it is not a precondition of `approved`, and no spec in the corpus makes an attestation one.

PR #244, opened 2026-09-17
docs(110): draft spec 110, replacing spec 071's stale acceptance

bartekus:
Files spec 110, `status: draft`. **Last** of the four follow-ons spec 106 §4 named; with it the audit's four red blocks are closed.

## What is red

`spec-spine verify 071` has failed at command 10 since spec 072 merged:

```
grep -qF 'this would update main' kit/settings.json
```

The shipped message reads `[push-gate] BLOCKED: this would update $def (repo: $root, branch: ${br:-unknown}). …`. Running 071's block one line at a time to the end, command 10 is the only red one; command 11 is green.

## After 072 the literal is the thing 071 §3.2 forbids

071 §3.2's requirement is:

> The refusal message MUST say what was actually refused, rather than restating the rule, and MUST name the tag push as allowed, so a maintainer who hits it can tell a correct refusal from a misfire.

The word `main` appears nowhere in it. 071 was written for a repository whose default branch is `main`, so the literal was a true rendering on the day it was written. 072 then made the protected branch a resolved value.

After 072 the literal is not merely stale. On a repository whose default is `master` or `trunk`, a message reading "this would update main" would name a branch the gate did not protect — which is exactly the misfire §3.2's sentence exists to let a maintainer detect. The assertion now points away from the requirement it was written for.

072 declared `amends` on 046, whose §3.4 states the refusal in terms of the literal name, and owed 071 none: 071 states a property of the message and 072 preserved it. Same finding as spec 108 §3.6 records for 093 and 059.

## The message was checked in one copy where everything else is checked in three

071's other six greps run against `kit/settings.json`, `.claude/settings.json` and `crates/spec-spine-core/src/kit_embedded.rs`. The message got one line.

§1.3 is explicit that this is **not** the coverage hole it looks like: `kit_hooks.rs` asserts the hook bodies in `.claude/settings.json` are the ones `kit/settings.json` ships (spec 051 §3.6) and `scaffold.rs` asserts the embedded copy has not drifted, and the block runs both at commands 2 and 3. The extension to three copies is **consistency with the surrounding lines, not new coverage**, and D-4 records that the first draft claimed otherwise and was corrected by checking. A spec in a series about assertions that claim more than they establish cannot afford an argument that does the same.

A fourth copy, `.codex/hooks.json`, is deliberately excluded: it arrived with spec 081, long after 071 (D-5's reasoning).

## What this PR does

No code, and in particular the hook bodies are untouched: they are correct, and what was wrong was an assertion about them. 110 declares `amends` + `amends_verification` on 071 and carries the replacement block. Spec 071's file is not edited (spec 040 §3.1), and the replacement asserts it still carries the superseded literal.

Command 10 becomes two assertions per copy: that the message reads `BLOCKED: this would update $def`, and that `this would update main` appears **nowhere**. D-2 records why the negative is not redundant: two refusal messages can coexist in one `case`, and a hook that grew a resolved message while keeping a literal arm would pass the positive line while shipping two refusals that disagree about what is protected. It is the idiom 071's own §3.2 assertions already use.

D-1 records why the matched string keeps `$def` rather than becoming branch-agnostic: `grep -qF 'this would update '` also passes a hook that reverted to the literal.

## Failability evidence

Measured against copies of `kit/settings.json` with the message rewritten:

| Condition | Line |
|---|---|
| the pre-072 message, `this would update main` | the positive line fails |
| the same | the negative line fails |
| the tag-push sentence removed | that line fails |
| the files as they stand | all nine pass |

## Also out of scope, with reasons recorded

- **Asserting the message by running the gate** (D-3). `kit_hooks.rs` asserts verdicts, not text. Extending it is new test code under `crates/`, inside 071's territory, from a spec whose §2 says it establishes no code.
- **The `(repo: …, branch: …)` context** (D-5). It reads well and it is not 071's: `git log -S` puts its arrival in the 045/046/047 commit.

## Gate

```
spec-spine verify 071   -> passed (25 command(s))   # runs 110's block
spec-spine verify 110   -> passed (25 command(s))
```


bartekus:
Both notes are design notes and both are answered; neither needs a change, and the first is a property rather than a leak.

**The temp file surviving an assertion failure is deliberate.** `verify` stops at the first failing command, so the `rm -f` below a failed assertion does not run and the captured document stays on disk. That is the point of §3.4's redirect, and it was established under review on #239: when an assertion fails, the reader gets the complete, pretty-printed document at the path the line names, which is strictly more than an `assert …, d` argument dump would print. The next run's `>` truncates before reading, so a stale file can never be read as a fresh one, as the note observes.

**The `cargo test` summary pattern.** Accurate and, as the note says, pre-existing: it is spec 106 D-7's idiom. The alternative is a bare filtered run, which exits 0 on zero matches (measured there: `0 passed; 31 filtered out`) and would leave the line green while asserting nothing. A dependency on an external format that fails closed beats an assertion that cannot fail.

With this, the four blocks spec 106 §4 named are closed: 056 by 107, 059 by 108, 060 by 109, and 071 here.

PR #245, opened 2026-09-17
chore(109): ratify (draft -> approved, pending -> complete)

bartekus:
Flips spec 109 from `draft` to `approved` and `pending` to `complete`.

109 is a no-build spec: it replaces spec 060's stale `## Verification` block through `amends_verification` (spec 103 §3.2) and changes no code, so both lifecycle axes move together. Same shape as specs 105 (#236), 106 (#238), 107 (#241) and 108 (#243).

Two things reviews of the earlier PRs in this series asked about, answered once here:

- **Nothing in CI runs `spec-spine verify`.** AGENTS.md keeps it outside the gate chain deliberately, because the chain runs against PR branches whose contents are, in the general case, a stranger's. That is why spec 093 could change a document shape and leave spec 060's block red for releases with every gate green, which is the crossing 109 repairs.
- **`draft` does not suppress the substitution.** Spec 103 §3.2 excludes `superseded` and `retired` only, and 109 was verified at `draft` on #242 before this flip: `verify 060` ran 109's 22 commands, not 060's 11.

Evidence on this branch:

```
spec-spine verify 060   -> passed (22 command(s))   # runs 109's block
spec-spine verify 109   -> passed (22 command(s))
```

Before 109, `verify 060` was `FAILED at command 8`. Only 109's own registry and index shards move.


bartekus:
The finding asks for confirmation that the artifact pre-exists, and it does: it is spec 109's own `## Verification` block, which became spec 060's acceptance when #242 merged at 0f4910a, this branch's base.

```
$ spec-spine verify 060   # runs 109's block, not 060's
verify: 060-plan-answers-the-whole-question: passed (22 command(s))
$ spec-spine verify 109
verify: 109-the-answer-is-a-member-not-the-document: passed (22 command(s))
```

060's own block is 11 commands; 22 is 109's. Before 109 the first command was `FAILED at command 8`. There is no code to point at because 109 ships none: what was broken was an assertion, and the repair is an assertion.

PR #246, opened 2026-09-17
chore(110): ratify (draft -> approved, pending -> complete)

bartekus:
Flips spec 110 from `draft` to `approved` and `pending` to `complete`. This is the last PR of the acceptance-block audit spec 103 §4 named and specs 105 and 106 carried.

110 is a no-build spec: it replaces spec 071's stale `## Verification` block through `amends_verification` (spec 103 §3.2) and changes no code, so both lifecycle axes move together. Same shape as 105 (#236), 106 (#238), 107 (#241), 108 (#243) and 109 (#245).

Evidence on this branch:

```
spec-spine verify 071   -> passed (25 command(s))   # runs 110's block
spec-spine verify 110   -> passed (25 command(s))
```

Before 110, `verify 071` was `FAILED at command 10`.

## The audit, closed

| Red block | Over-assertion | Moved by | Repaired by |
|---|---|---|---|
| 050 cmd 5/7/8 | whole stdout, bare array, version pin | 057, 093, 088 | 106 |
| 056 cmd 5 | `schemaVersion == "0.3.0"` | 088 | 107 |
| 059 cmd 4 | exact key set `{orphaned, inFlight}` | 093 | 108 |
| 060 cmd 8 | whole document `== {id, title}` | 093 | 109 |
| 071 cmd 10 | literal `this would update main` | 072 | 110 |

098's two defects were 105's. Four of the five repairs correct assertions that were **stricter than their rules**; 109 is the one where the rule itself was amended by the book (093 carried the `amends` edge) and only the acceptance was out of reach.

Only 110's own registry and index shards move.


bartekus:
The one hard finding rests on lifecycle states this corpus does not have, so there is nothing to bypass.

**`status` has exactly four values.** `crates/spec-spine-types/src/frontmatter.rs:28`:

```rust
pub enum Status {
    Draft,
    Approved,
    Superseded,
    Retired,
}
```

There is no `planned` status and no `ready` status, so "add the missing `planned` and `ready` transition commits" is not a change anyone could make: the enum is `#[serde(rename_all = "lowercase")]` over those four, and either word in that field is a frontmatter parse failure (exit 3), not a state. `draft -> approved` is not a leap past anything; it is the only forward transition the axis has.

`planned` does exist in this corpus, but it is a per-unit boolean on an authority unit (spec 076), declaring territory before the code is written. It is not a lifecycle value, and 110 claims no units.

**The transition is the one AGENTS.md step 6 describes.** "After merge, a second PR flips `status: draft` to `approved` (the ratify PR), and the corpus count moves." This is that PR. The second axis moves with it because 110 ships no code: `implementation: complete` on a no-build spec means its artifact exists and works, and the artifact is its `## Verification` block, which became spec 071's acceptance when #244 merged at d2db459 — this branch's base.

```
$ spec-spine verify 071   # runs 110's block, not 071's
verify: 071-a-tag-push-is-not-a-push-to-main: passed (25 command(s))
```

Before 110 that was `FAILED at command 10`. 071's own block is 11 commands; 25 is 110's. Specs 105, 106, 107, 108 and 109 all ratified exactly this way.

PR #249, opened 2026-09-17
feat(111): teach the whole frontmatter grammar in the spec template

bartekus:
Closes the one item of the acceptance audit left standing after specs 106 to 110.

## The gap

`standards/spec/templates/spec-template.md` is what `/spec` copies to start every spec, and it is the only place in the corpus where the frontmatter grammar is written down for an author. Specs 106, 107, 108, 109 and 110 each carry the same sentence in their own section 4:

> `standards/spec/templates/spec-template.md` does not carry the key spec 103 added. 106 §4 names it as a real gap in territory neither spec owns.

Each correctly declined to fix it. A spec may claim territory, and none of them claimed this file:

```
$ spec-spine index owner standards/spec/templates/spec-template.md
standards/spec/templates/spec-template.md
  (no spec owns this path)
```

The file was nonetheless hashed (it matches `standards/**/*.md` in `[index] extra_hashed_inputs`), so its bytes were folded into the global-inputs scalar and signed by every attestation. A governed byte with no governor.

## The gap is five keys wide, not one

Measured against `spec_spine_types::KNOWN_KEYS` (29 entries) at `2a3e0e7`, five were absent:

| Key | Status in the tool |
| --- | --- |
| `code_aliases` | parses, compiles to the registry, no verb reads it |
| `feature_branch` | parses, compiles to the registry, no verb reads it |
| `amends_verification` | spec 103: `verify` resolves through it |
| `amendment_record` | `couple.rs` adds it to the amended spec.md's owner set |
| `origin` | the retroactive bootstrap marker; spec 000 declares it |

Two nested forms were missing too, and neither is obscure: `nature` on an `extends` edge, which **96 of the 111 specs** in this corpus write, and `planned: true` on a unit (spec 076), written by twelve. Neither `contract.md` nor `constitution.md` mentions any of the seven.

## What this does

- **Claims the template.** `index owner` now answers for it.
- **Documents `amends_verification`** with the guidance that makes it usable and is not derivable from the syntax: when to reach for it, the `amends` entry every entry requires and `V-018`, the `V-019` fork and `V-020` cycle refusals, chain resolution past a `superseded` amender, the attribution line `verify` prints, carrying the amended block in full rather than as a patch, and asserting that the amended file still carries the superseded form. That last is the practice specs 107 to 110 each adopted and none could write down anywhere an author would find it.
- **Documents the other four keys.** `code_aliases` and `feature_branch` are marked **inert**, because documenting them silently as ordinary authoring keys would teach an author to write a key nothing reads.
- **Adds the nested forms**: all six unit granularities, the trailing-slash subtree, `nature`, `planned: true`, and the `paths:` sugar.
- **Adds `## 5. Resolved decisions` and a `## Verification` fence**, the two sections every spec since 048 carries and the template never mentioned. A placeholder fence reports `not-declared`, not a vacuous pass (`verify.rs:169`), so no new empty-but-green block enters the corpus.

## Why the check is Rust and not only a verify line

The template fell behind because nothing connects it to the grammar: spec 103 added a key to `KNOWN_KEYS`, the DTO, `verify`'s resolution and two test suites, and the gate stayed green because nothing reads the template. Adding one key by hand leaves that mechanism exactly as it was.

`crates/spec-spine-types/tests/dogfood.rs` now iterates `KNOWN_KEYS` itself, not a list transcribed into the test (a third copy of the grammar would go stale the same way). `cargo test --workspace` is in the gate chain and `verify` deliberately is not, so the next key added to the grammar cannot reach `main` without the document that teaches it.

**Fail-first evidence.** Against the parent template the new test reports, naming the five:

```
standards/spec/templates/spec-template.md documents 24/29 frontmatter keys;
missing: ["code_aliases", "feature_branch", "amends_verification", "amendment_record", "origin"]
```

## Gate

| verb | result |
| --- | --- |
| `check --fail-on-unresolved --fail-on-warn` | registry fresh, index fresh |
| `lint --fail-on-warn` | 0 errors, 0 warnings, 0 info |
| `index coverage --fail-on-untraced` | 95/95 claimed, 0 unclaimed |
| `couple` | OK, 3 paths checked, no drift |
| `verify 111` | passed (33 commands) |
| stack | `cargo test --workspace` + `--no-default-features`, clippy `-D warnings`, `fmt --check` all clean |

The 117 derived shards in this diff are the expected restale: the template matches `standards/**/*.md` in `extra_hashed_inputs`, so editing it folds into the global-inputs scalar.

## Deliberately not here

The adopter scaffold template (`scaffold.rs::spec_template()`) is a separate, minimal starter for `spec-spine init`; it is spec 006's territory and a different audience. The contract and constitution are owned by 040/043/066. Giving `code_aliases` and `feature_branch` meaning stays the deliberate decision `docs/design/02-agentic-builder-substrate.md` says it should be.

Filed `status: draft`. Approval is a human flip in a follow-up ratify PR.


bartekus:
## Review disposition

Four findings. One applied, three rejected with evidence.

### Applied: vacuous `KNOWN_KEYS` guard

Correct, and it is the shape this spec exists to remove. Fixed in `970edf3`, but **not** with the proposed `assert!(KNOWN_KEYS.len() >= 29)`: that is a literal pin on a list designed to grow, and it goes red on a legitimate retirement of any key, which is exactly the over-assertion specs 107 and 108 were filed to repair. The guard instead names the five keys §1.2 measured as absent and asserts each is still in `KNOWN_KEYS`. Same line count, cannot fire on an unrelated change, and it says what it protects. Requirement recorded in §3.5, reasoning in D-6, acceptance line added (34 commands, was 33).

Worth noting the vacuous case was already unreachable in silence: `bootstrap_spec_000_parses`, two functions up in the same file, asserts `fm.extra_frontmatter.is_empty()`, which an emptied `KNOWN_KEYS` would break loudly. But a tripwire whose non-vacuity depends on a second test elsewhere is not one a reader can check, so the finding stands on its own terms.

### Rejected: `registry show 111` needs a preceding `compile`

The premise is wrong and the proposed remedy would introduce a defect.

The registry shards are **committed** in this repository (self-governance): `git ls-files .derived/spec-registry/by-spec/ | wc -l` is 112, spec 111's shard among them. `registry show 111` therefore answers on any checkout with no compile, which is why `verify 111` passes 34/34.

More importantly, adding `spec-spine compile` to a `## Verification` block would make the block **write to the tree it is judging**. `.claude/rules/orchestrator-rules.md` and the `/prime` protocol both forbid that ("a gate must never repair the tree it is judging"), and it would mask precisely the staleness these lines should expose: if the committed shard were stale, they are supposed to fail. The corpus already reflects this. Every verify block that touches compile uses the non-writing `compile --check` form (specs 052, 055, 057, 063), never a bare `compile`.

### Rejected: single `# ` strip in the `documents` closure

The finding concedes both halves itself: "No current breakage" and "the test goes red, which is the right failure mode." Nothing to act on.

### Confirmed, no change: coupling-gate compliance

The structural assumption named is spec 102, and it is measured rather than assumed: `couple --base origin/main --head HEAD` returned `OK: 3 path(s) checked, no drift` locally at the commit, and the `self-governance (compile · index · lint · couple)` job passed on CI for `37cb06e`. The `extends` edge on spec 000's `dogfood.rs` is read from the in-flight spec.md, as spec 102 intends.

bartekus:
## Second review pass: one new claim, and it does not hold

This pass is mostly confirmations (coupling coverage, the `depends_on` ids, the non-vacuity guard, the pass-count grep, D-5's `kind` analysis, all marked correct) plus a restatement of the single-`# `-strip item already answered, which this pass itself labels "Not a finding against this PR". One claim is new.

### Rejected: "D-4 cites the wrong file"

The claim is that `.claude/rules/adversarial-prompt-refusal.md` "governs LLM adversarial-prompt handling, not spec coupling semantics" and that D-4's citation points at an unrelated file.

The file's closing sentence, verbatim:

> If the code needs to touch a unit another spec owns, declare an `extends` edge naming that spec and unit; that amends nobody.

That is precisely and only what D-4 cites it for. The rule is the repository's coherence guard and its whole subject is coupling semantics: what may be edited mid-build, that a waiver is a human instrument, and which edge to declare when code must touch another spec's unit. The suggested substitute, spec 052, names the crossing in a `C-001` footer; it does not state the authoring rule D-4 is invoking. No change.

### Non-finding: "`KNOWN_KEYS` is imported but not shown being exported"

Self-hedged in the comment, and answerable from the tree: `crates/spec-spine-types/src/lib.rs:73` re-exports it. `build · test · clippy` also passed on this head sha, which it could not have if the test did not compile. The conditional worry about `implementation: complete` therefore does not arise.

### Non-finding: `python3` with no fallback

Labelled "Not a blocker" in the comment. It is the corpus convention (specs 103 and 107 to 110 all shell out to `python3` in their blocks) and `verify 111` passes 34/34 on the runner. Changing it here would make this block differ from every neighbouring one for no gain.

Remediation stops here: this pass named no new code path, which is the point at which further rounds restate rather than find.

PR #251, opened 2026-09-17
feat(112): sweep the corpus's acceptance from a maintainer's hand

bartekus:
Closes the gap spec 105 §4 named by name: *"the remedy is a periodic sweep a
maintainer runs, not a gate step; naming where that sweep belongs is its own
work."* This is that work.

## The gap

`verify` is the one verb that executes what the corpus declares, so it sits
outside the gate chain deliberately and nothing in CI runs it. A
`## Verification` block is therefore run once, by the session that writes it,
and then never again. Seven blocks went red against a moving corpus and every
one was found by hand: specs 103, 105, 106, 107, 108, 109 and 110 each exist to
repair one.

## What this adds

`scripts/verify-sweep.sh`: a maintainer-invoked sweep that runs the whole
corpus's declared acceptance against one already-merged revision in a detached
worktree. It is **not** a CLI subcommand, a CI job, a hook, or a step in any
skill's gate floor, and nothing about `verify`, the gate chain or any schema
changes. Spec 049 §4 declined `verify --all` on two grounds: the loop belongs
in a caller, and the `not-declared` policy needed grounds it did not have. This
spec supplies the second and honours the first.

**Every spec gets one of five outcomes, and only two are success:**

| outcome | success |
|---|---|
| `passed` | yes |
| `exempt` | yes |
| `failed` | no |
| `not-declared` | no |
| `not-run` | no |

Missing acceptance never counts as passing. `not-declared` is reported only
when `verify --plan` succeeded and returned nothing; a spec whose plan could
not be read is `not-run`, because "declares no acceptance" and "we could not
find out" are different facts.

**The 000-047 population is settled as a closed ledger.** Those 48 specs are
exactly the ones filed before `verify` existed (spec 049 built the verb; spec
048, filed immediately before it, is the first spec in the corpus to carry a
block). They are carried as an enumerated list with three refusals enforced
before any block runs: an entry at or above ordinal 48 is refused, so no future
spec can enter it; an entry whose spec declares acceptance is refused as a
stale exemption, so the ledger only shrinks; a dangling entry is refused. The
debt is tracked and reviewable in `git log -p`, and it cannot grow.

**The trust boundary is enforced rather than documented.** The sweep refuses a
revision that is not an ancestor of the trusted ref (default `origin/HEAD`), so
it cannot be pointed at a PR branch by accident, and it records the ref it
used. Arbitrary corpus verification stays out of the PR gate in both
directions.

**Isolation and failure behaviour.** Detached worktree, removed at exit;
per-spec `TMPDIR`, so one block's fixture residue cannot decide another's
outcome; the run never stops at a failure; a block that dirties the worktree is
recorded and the tree restored before the next spec; a per-spec limit enforced
in-script (not via `timeout(1)`, which stock macOS does not ship) so a hang
costs one result rather than the run; the report is refused inside the
repository under test.

## First run, measured

Against `4e114fd` (merged tip of `main`), binary built from that revision:

```
passed=64  failed=0  not-declared=0  exempt=48  not-run=0
112 specs in 6m24s, no block left the worktree dirty
```

The corpus is clean, which is the expected reading: specs 105-111 retired the
known debt over the last two days. What changed is the cost of knowing it.
No follow-up repair work was discovered, so nothing here edits an approved
spec, and this PR carries no waiver.

## Acceptance

`spec-spine verify 112`: **passed (56 commands)**. The block builds an
eight-spec fixture corpus that produces all five outcomes in one run, and
exercises the four ledger refusals, the trust boundary, the containment guard,
the run-directory guard, the per-spec limit and the dirty-tree restoration.

Fail-first was measured, not asserted: every line was run in a worktree of the
parent commit `4e114fd`. Lines 1-14 build the fixture and legitimately succeed
there; line 15, the first sweep invocation, is red. Three lines are
deliberately not fail-first and §"Verification" says which and why.

## Gate

`check` fresh/fresh · `lint` 0/0/0 · `couple` 3 paths, no drift ·
`index coverage` 95/95 · `cargo test --workspace` 38 suites green ·
`clippy -D warnings` · `fmt --check` · `shellcheck` clean.

Spec 112 is `draft`; ratification is a separate human PR.


bartekus:
## Review disposition

Five automated passes. Six findings applied over four rounds, four judged non-findings with the evidence for each.

**Applied**

| pass | finding | fix |
|---|---|---|
| 1 | a plan-failure `not-run` row cited a log that was never written | the report cites a log only when the file exists; acceptance asserts `007-ghost` cites none and every cited path resolves |
| 2 | the containment guard `mkdir -p`'d the parent before refusing, so `--out <repo>/new/run` created `<repo>/new` inside the tree it protects | canonicalize by walking up to the deepest existing ancestor, creating nothing; acceptance covers the case the old `--out <repo>/inside` line could not reach |
| 2 | §1.4 was ordered ahead of §1.3 | swapped |
| 3 | the fallback from a process-group kill to a process kill silently did something weaker | the fallback stays and now writes a line into that spec's log, and §3.6 says so, so a degraded run cannot read as a clean one |
| 3 | a flat `sleep 2` between SIGTERM and SIGKILL | polls instead, so the grace is proportional |
| 4 | `--only 001,001` ran the spec twice and counted it twice | selection takes each spec once; §3.5 says so and the acceptance asserts it |
| 4 | `run_limited` returned 124 unconditionally, so a block that finished on its own at the boundary was reported `not-run` despite reaching a verdict | the child's own status is kept when there is one |

**Not findings, measured rather than argued**

- *`--only 12` does not match `012-...`.* Correct, and deliberate: `spec-spine` itself answers `not found: spec '49'`, so a sweep that padded would be the one place in the project where a short id means something wider. Fixed the wording in §3.5, `--help` and the refusal, which now names the form it wants.
- *`--only '0[1'` causes a case-pattern parse error.* It does not. `"$want"-*` expands the variable inside quotes, so its contents match literally; run against both a matching and a non-matching id, no error and no match.
- *`007-ghost` relies on an unverified tool guarantee.* It is verified: the acceptance runs the real binary and asserts `exitCode != 0` with `--plan` named in the failure, so the line goes red if `verify --plan` ever answers 0 for a spec the registry lists and the tree does not carry.
- *A relative `--out` is silently truncated.* It is not. `--out relout/deep/run` resolves to `$PWD/relout/deep/run` and the report records the absolute path; the walk-up handles `.` and `/` as terminals. Tested directly.

**Evidence on the final sha**

`spec-spine verify 112`: passed, 66 commands. Gate: `check` fresh/fresh, `lint` 0/0/0, `couple` 3 paths no drift, `coverage` 95/95, `cargo test --workspace` 38 suites, `clippy -D warnings`, `fmt --check`, `shellcheck` clean.

Full sweep re-run from the shipped script against `4e114fd`: `passed=64 failed=0 not-declared=0 exempt=48 not-run=0`, 112 specs in 5m18s, no block left the worktree dirty.

PR #254, opened 2026-09-18
feat(113): the scaffolded protocol is the gate the kit ships

bartekus:
## Summary

`spec-spine init --with-kit` writes two documents into one tree and they disagreed about the same gate. The `AGENTS.md` generated from the string literal in `scaffold.rs` named `spec-spine index check --fail-on-unresolved`, `couple --base origin/main`, an unconditional `index coverage --fail-on-untraced`, and a startup read of `spec-spine compile --check`. `kit/AGENTS.md`, which the same command copies into the same tree, has named `spec-spine check`, a resolved base ref, and a conditional ownership assertion since spec 075. Every invocation on the left still worked, so nothing was broken; an adopter simply got the older gate, told in two voices by one command.

The generated gate list is now the kit's, step for step, including both conditional lines and the sentence that carries `$SPEC_SPINE_DEFAULT_BRANCH`'s actual scope (the push gate and `Makefile`, per 113 D-3). The startup freshness read is the composed `check`. The paths stay the adopter's: 065 section 3.1's config-awareness is asserted against a non-default `specs_dir`, whole-path, with the negative half that a hard-coding generator would fail.

`crates/spec-spine-core/tests/kit_gate.rs` gains the assertion section 1.2 says did not exist. `gate_steps` parses a fenced gate list out of either document and keeps each step's conditional flag; `the_scaffolded_protocol_gate_matches_the_kits` compares the generated list against `kit/AGENTS.md`'s. Editing either list alone now fails a test rather than shipping.

Spec 065's acceptance pinned the spelling `compile --check` rather than the property 065 section 3.1 states, so this spec replaces that block in full through `amends_verification` (spec 103) and does not edit 065. Filed and built in one change: 103's resolver treats a draft holder as live, so merged as a filing alone this would leave 065's acceptance red on the default branch (113 section 3.6, D-5).

Specs 114 to 117 are a separate wave and are deliberately not in this PR.

## Testing

Gate as `AGENTS.md` lists it, all exit 0 on the committed tree:

- `spec-spine compile`, `spec-spine index`
- `spec-spine check --fail-on-unresolved --fail-on-warn`: spec-registry fresh, codebase-index fresh
- `spec-spine lint --fail-on-warn`: 0 errors, 0 warnings
- `spec-spine index coverage --fail-on-untraced`: 95/95 claimed, 0 unclaimed
- `spec-spine couple --base "$(git symbolic-ref ...)" --head HEAD`: 4 paths checked, no drift
- `cargo test --workspace --locked`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all --check`

Acceptance:

- `spec-spine verify 113`: passed, 33 commands
- `spec-spine verify 065`: resolves to this spec's block through `amends_verification` and passes, 33 commands

Fail-first evidence, measured against the parent's `scaffold.rs`:

- `the_scaffolded_protocol_gate_matches_the_kits` and `the_scaffolded_protocol_names_the_composed_reads` both fail at 7a7a8b6 and pass here
- the six greps section 3.5 tabulates all flip: `spec-spine check` absent to present, `spec-spine compile --check` present to absent, `symbolic-ref` and `require_ownership` absent to present, `couple --base origin/main` and `index check --fail-on-unresolved` present to absent
- section 3.3's three config-awareness lines are green at the parent and stay green, which is preservation rather than evidence; they are what would catch a generator that hard-coded paths while rewriting the gate
- the parser's own drop guard was proved to fire: a comment marker written without a space gives "7 fence line(s) name spec-spine and 6 parsed as steps"


bartekus:
Addressed in 18d8ee5. Grading each finding:

**Close-fence detection is fragile: accepted, fixed.** The finding is right and it is sharper than stated: because both `named` and `steps` are counted over the same truncated slice, the drop guard added beside the parser could not have caught it. The close is now the first fence that begins a line, allowing leading whitespace, since `kit/AGENTS.md` carries the block indented inside a numbered list. Probed: an inline backtick run inside the fence is now read past instead of truncating the list silently.

**Double-space comment split: accepted, fixed.** Split on one space rather than two. `.trim()` removes the second where the generator writes it, so both spacings parse to the same command. The concrete risk was the false positive, not the false negative the comment describes: a single-spaced condition in the hand-maintained `kit/AGENTS.md` would have folded the prose into the command and failed parity against a correct scaffold. Probed both directions.

**`crates/spec-spine-cli/tests/init.rs` claimed but unchanged: no change needed.** The coupling gate refuses a changed path no spec claims, not a claim whose path did not change, so this is over-declaration rather than a violation, and your reading of the intent is the right one. The `--with-kit` assertions in `init.rs` assert that the file lands, not what it says, so they hold unmodified: the acceptance runs `cargo test -p spec-spine-cli --test init --locked` and it passes at both the parent and here. Section 2 declares standing territory for the unit; narrowing it mid-build would be editing what the spec claims to match what the diff happened to touch.

Gate green locally after the fix: compile, index, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (4 paths, no drift), `cargo test --workspace`, `clippy -D warnings`, `fmt --check`. `verify 113` and `verify 065` both pass, 33 commands each.

bartekus:
Second pass addressed in 9b15d7d.

**Drop guard fires on prose mentions: accepted, fixed.** `named` counted any fence line containing the string, so a comment mentioning the binary would raise the guard's failure against a well-formed fence. It now counts lines that read as an invocation under a lenient reading of the comment marker, where the parser applies a strict one, which is the gap actually worth guarding: `#spec-spine check` without the space still counts here and still parses to nothing, while prose is an invocation under neither reading. Probed both directions.

**`crates/spec-spine-cli/tests/init.rs`: confirmed, no change needed.** The specific worry is checkable and checks out. `init.rs` makes no assertion about the content of the scaffolded protocol: it asserts `AGENTS.md` is in the emitted file list (line 174), that it lands on disk (213), and that an existing one is preserved rather than overwritten (232, 246, 265). Nothing there was obsoleted by inverting `tests/scaffold.rs`, which is where the gate-text assertions live. The acceptance runs `cargo test -p spec-spine-cli --test init --locked` and it passes unchanged at the parent and here. The `extends` entry is standing territory.

**`index check` in the verification block: no change, deliberately.** Correct that it is not a defect. Leaving it is also not an oversight. That line is spec 065 section 3.4's, carried forward under spec 103 section 3.5's requirement to reproduce the amended block in full, and section 3.5 of this spec moves exactly one assertion in it. `index check` is a primitive this spec does not deprecate; what it deprecates is naming the primitive in a **protocol** where the composed verb belongs, which is the four lines section 4 scopes the change to. Rewriting a second acceptance command because it reads oddly would be widening the replacement past what the spec declares.

Gate green locally: compile, index, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (4 paths, no drift), `cargo test --workspace`, `clippy -D warnings`, `fmt --check`. `verify 113` and `verify 065` both pass, 33 commands each.

bartekus:
Third pass: no change. Both remediation rounds are spent, and this pass names no new code path.

**`crates/spec-spine-cli/tests/init.rs`:** third restatement of a question answered twice with file and line evidence. The specific hypothesis is checkable and false: `init.rs` contains no `contains(...)` assertion against the protocol's text at all. Its five references to the file are the emitted-file-list entry (174), the on-disk existence check (213), and the preserve-an-existing-one case (232, 246, 265). The gate-text assertions live in `tests/scaffold.rs`, which this PR inverts. `cargo test -p spec-spine-cli --test init --locked` passes at the parent and here, which is not a claim, it is in the acceptance and in CI.

**Close fence at position 0:** not a defect. If `body` began with a fence, the block would be empty and position 0 would be the close, so identifying it there is correct rather than mistaken. `body` starts immediately after the literal ` ```sh `, which is followed by a newline in both documents, so the case does not arise; were it to, `assert!(!kit.is_empty())` refuses an empty list before parity is ever compared.

**Exit 3 for `spec-spine check`:** verified rather than assumed, since the point is fair. A tree carrying a malformed `spec-spine.toml` gives `config error: TOML parse error ...` and exits 3. The generated protocol documents a code the binary returns.

The two confirmations in the minor notes match what was measured here: the negative `specs/` assertion is clean against a `corpus/` layout, and the resolved base ref contains no ` #`.

PR #255, opened 2026-09-18
chore(113): ratify (draft -> approved)

bartekus:
## Summary

Spec 113 shipped in PR #254 (`beaad2f`). This is its ratification PR: one
frontmatter line, `status: draft` -> `status: approved`, plus the two derived
shards that line restales.

`implementation: complete` is unchanged. No requirement, no acceptance
criterion, no implementation file, and no predecessor spec is touched.

## Why now

113 made `spec-spine init` generate an `AGENTS.md` whose gate chain is the
kit's, step for step, so one command no longer writes two disagreeing gates
into one tree. It also declared spec 065's acceptance replaced, via
`amends_verification`, rather than editing 065's file: 065's assertion grepped
the scaffolded protocol for `spec-spine compile --check`, an assertion narrower
than the rule it stood for, which the change falsified.

## Validation

Gate as `AGENTS.md` lists it, all green on the binary in-tree (`0.20.0`):

| step | result |
|---|---|
| `check --fail-on-unresolved --fail-on-warn` | 0 (registry fresh, index fresh) |
| `lint --fail-on-warn` | 0 errors, 0 warnings, 0 info |
| `index coverage --fail-on-untraced` | 95/95 claimed (100.0%), 0 unclaimed |
| `couple --base origin/main --head HEAD --include-uncommitted` | OK, 1 path checked, no drift |
| `cargo test --workspace --locked` | pass |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | pass |
| `cargo fmt --all --check` | pass |

Acceptance:

- `spec-spine verify 113`: passed, 33 commands.
- `spec-spine verify 065`: reports `acceptance amended by 113-the-scaffolded-protocol-is-the-gate-the-kit-ships (spec 040); its block is the one that runs`, then passes the same 33 commands. The replacement resolves and holds.

Corpus after this merge: 114 specs, 114 approved, 0 draft.


bartekus:
Both automated findings were checked against the corpus and neither is supported.

**Finding 1 (attestation record):** spec 042 §3.3 is titled *"It is not committed, and that is a decision"*, and `.gitignore:30` ignores `.derived/attestation/`. The per-spec attestation is an on-demand artifact by design, so no attestation record can ever appear in a ratify diff, and no spec requires one to accompany a status flip. The verified evidence for this promotion is `spec-spine verify 113` (passed, 33 commands) and the full gate, both recorded in the PR body. Spec 049 governs that verb, not a committed record.

**Finding 2 (no implementation in diff):** case (a), which the finding itself allows. The implementation landed in #254 at `beaad2f`, which is the head of this PR's base branch. Spec 041 holds that a `complete` spec is not in flight; it does not bear on a draft-to-approved flip whose implementation is already merged, and the gate's `check --fail-on-unresolved` passed.

No change made.

PR #260, opened 2026-09-19
feat(118): the verdict is the only thing on stdout

bartekus:
Spec 118, filed and built in one change.

## The defect

Spec 037 §3.1: "Under `--json` a verb MUST write exactly one JSON object to
stdout." `verify --json` did not. It spawned each acceptance command with
`Command::status()`, which inherits the parent's stdout, so every byte the
command printed landed on the stream the envelope was about to be written to,
ahead of it.

The failure is quiet in the worst way. Measured on `5f95613` with
`spec-spine 0.20.0`, against an isolated fixture whose block is one command
printing one line:

```
$ spec-spine --repo "$F" verify 001-noisy --json > out 2> err; echo $?
0
$ head -1 out
CHILD-STDOUT-LINE
$ python3 -c 'import json; json.load(open("out"))'
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The verdict is correct, `ok` agrees with the exit code, the report carries
every field spec 049 §3.4 lists, and the process exits 0. The only thing wrong
is the channel, so a consumer reading the exit code sees a pass and a consumer
reading the document sees a parse error on a successful run. The obvious
diagnosis, that `verify` emitted malformed JSON, is wrong.

Spec 049 §3.5 also requires the CLI to "echo each command and its exit code, so
that a CI log reads as a transcript". That clause names no channel, and both
echo sites sat inside `if !json`, so a machine-readable run had no transcript at
all: the mode a CI log is most likely to be produced in.

## Why the tests could not see it

`verify_json_is_a_verdict_envelope_that_agrees_with_the_exit_code` already
parsed stdout with `serde_json::from_slice` and already carried the message
`"stdout is one JSON envelope"`. It passed, because every command in its
fixtures is `true`, `exit 7`, or a `verify:browser` fence that never runs. Not
one of them writes a byte.

## The correction

Under `--json`, each child is spawned with its stdout and stderr piped, and both
are forwarded to the parent's stderr as they arrive. The transcript joins them
there; without the flag every byte goes exactly where it goes today.

- **Streamed, not accumulated.** A pipe buffer is finite, so a forwarder that
  waited for the child to exit would deadlock on the first command verbose
  enough to fill one, and `cargo test` over a workspace is that command. The
  child's stdout gets a thread of its own and its stderr is drained on the
  calling one, so neither pipe can block the other. `io::copy` streams through a
  fixed buffer, so a command may print megabytes without the verb growing with
  it.
- **Diverted at the spawn, not filtered at the write.** That is what makes the
  guarantee total: the report path and the error path are covered by the same
  fact, that the child never holds the stdout descriptor.
- **No schema change.** No field is added, so `VERDICT_SCHEMA_VERSION` does not
  move. The child's output is not put in the report, where it would make the
  envelope grow without bound with the output of `cargo test`.
- **The cost, stated.** The two pipes are buffered independently, so §3.2
  promises no ordering between them. A caller that needs the command's own
  interleaving can merge the streams in the command with `2>&1`.
- **Preserved, and asserted rather than assured:** prose-mode channels,
  execution through `sh -c` from the repository root in plan order,
  stop-on-first-failure, the inherited environment, `SPEC_SPINE_VERIFY_STACK`
  and the `R-001` refusal, every report field, the exit-code mapping, and
  `--plan --json` as a non-executing operation. Subprocess execution stays in
  the CLI, so the seam spec 049 §3.1 draws is untouched.

## Authority

`extends` on `cmd_verify.rs` (spec 049) and on `tests/cli.rs` (spec 001);
`establishes` the new `tests/verify_streams.rs`. **No `amends` edge and no
`amends_verification`.** Spec 049 §3.5's requirement is not withdrawn: it
requires the echo and names no channel, so §3.3 names the channel per mode,
which satisfies it in both modes instead of dropping it in one. Spec 049's file
is not edited, and its own acceptance still passes unchanged. Spec 103 §3.4's
"acceptance amended by" notice is deliberately left where it is, in the prose
mode only, with the reasoning in §4.

## Evidence

Fail-first, measured in a detached worktree at `5f95613` with a release binary
built there:

| acceptance line | at `5f95613` | at this head |
|---|---|---|
| passing command, JSON channels | **1** (`JSONDecodeError` at char 0) | 0 |
| failing command, JSON channels | **1** (same) | 0 |
| `--plan --json` runs nothing | 0 | 0 |
| `R-001` error envelope is the whole of stdout | 0 | 0 |
| prose-mode channels unchanged | 0 | 0 |
| the engine spawns nothing | 0 | 0 |

The last four are preservation, not evidence, and the spec says so line by line.
`tests/verify_streams.rs`: 3 of its 6 cases red against the parent's
`cmd_verify.rs`, 6 of 6 green here. `tests/cli.rs`'s envelope case: red against
the parent once its fixture writes.

Two assertions in the first draft of the suite were vacuous and were caught by
measuring rather than by reading: a literal marker appears in the transcript on
stderr and in the envelope's `failure.command` on stdout, so
`stderr.contains("OUT-A")` passed on the echo of the command and
`!stdout.contains("OUT-A")` failed on the envelope's own payload. Every marker
is now assembled by the child at run time (`printf 'O%sT-A\n' U`), in the test
file and in the `## Verification` block alike.

Gate, in the `AGENTS.md` order, all exit 0: `compile`, `index`,
`check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`,
`index coverage --fail-on-untraced` (95/95 claimed),
`couple` (4 paths, no drift), `cargo test --workspace --locked`,
`cargo clippy --workspace --all-targets --locked -- -D warnings`,
`cargo fmt --all --check`.

`spec-spine verify 118` passes (9 commands). Affected acceptance: 049 passes
(9), 103 passes (21), 051 passes (13), 106 passes (21); 001, 035 and 037 are
`not-declared`.

## Lifecycle

`status: draft`, `implementation: complete`. Ratification is a separate PR and a
human flip.


bartekus:
Review findings, graded against source. One accepted and fixed in `cbc92e1`; two are false positives, with the evidence below.

### 3. Pump thread not joined on `child.wait()` error — ACCEPTED, fixed

Correct, and correctly scoped as low severity: the thread drains to EOF once the child is dropped, so there is no deadlock. What it did cost is exactly what the finding names: the ordering the comment beside it claims held on every path except that one, and a panic inside the thread was discarded rather than observed. `cbc92e1` holds the wait's result, joins, and propagates afterwards, so both paths join.

### 1. Ordinal gap violates spec 053 (monotonicity) — FALSE POSITIVE

Spec 053 is `L-007`, and its requirement is that a `depends_on` entry name a **lower** ordinal than the spec declaring it. From its own summary: "that a `depends_on` entry names a lower ordinal than the spec declaring it". It says nothing about gaps in the filing sequence, and there is no rule anywhere in the corpus that ordinals be contiguous. Spec 118's `depends_on` is `035`, `037`, `049`, every one of them backward, which is the property 053 actually checks. `lint --fail-on-warn` exits 0 (and `require_ordinal_monotonic_depends_on` is off in `spec-spine.toml` besides, per 053's own default-off decision).

The gap is deliberate. 115, 116 and 117 are reserved identities carrying draft spec directories on the wave branch `113-the-harness-delivers-what-it-documents` at `487bbd9`; renumbering them to close a gap would collide with work in flight, and renumbering is a human decision in any case.

### 2. Wrong owning spec in the `extends` annotation for `tests/cli.rs` — FALSE POSITIVE

`crates/spec-spine-cli/tests/cli.rs` has no establisher. Spec 001 is its **floor** owner, through the crate's package manifest, which `index owner` answers directly:

```
$ spec-spine index owner crates/spec-spine-cli/tests/cli.rs
crates/spec-spine-cli/tests/cli.rs
  001-compile-registry    floor   crates/spec-spine-cli (package manifest)
  010-registry-query-projection-flags   unit   ... (extends)
  ...
  049-verify-declared-acceptance        unit   ... (extends)
```

Every spec that has ever touched this file, spec 049 included, carries it as an `extends` crossing naming **001**. Spec 049's own frontmatter line is `- { spec: "001-compile-registry", unit: "crates/spec-spine-cli/tests/cli.rs", nature: additive }`, which is the same line this PR is being asked to change. Crediting the establisher, or the floor where there is none, is the convention throughout the corpus.

The inference that "the gate will not see the 049 crossing" is also not how `couple` works. The gate asks whether an **owning spec of the changed path** changed in the same diff, and an extends-carried unit is a first-class claim regardless of which spec the edge targets (`CLAUDE.md`, "How a file gets claimed"). Spec 118 owns this path by its own edge, `specs/118-.../spec.md` is in the diff, and `couple` reports `OK: 4 path(s) checked, no drift`.

bartekus:
Second pass, graded. One code finding accepted and fixed in `00ba7bb`; the two spec-compliance findings are false positives, and one of them is a restatement of a claim already answered above with `index owner` output.

### `if let Some` on `child.stderr` is unreachable — ACCEPTED, fixed

Graded as cosmetic in the review and taken anyway, because the branch is not harmless to leave: the only behaviour it could ever have is skipping the stderr drain, which is precisely the thing that lets `child.wait()` block on a full stderr pipe. `00ba7bb` takes the handle with `expect` like the stdout handle beside it, so there is no branch and no unreachability argument for the next reader to reconstruct.

### `extends` cites spec 001 for `tests/cli.rs` — FALSE POSITIVE (restated)

Already answered above with measured output. `crates/spec-spine-cli/tests/cli.rs` has **no establisher**; spec 001 is its floor owner through the crate's package manifest, which is what `spec-spine index owner` reports. Every spec that has ever crossed this file carries `{ spec: "001-compile-registry", unit: "crates/spec-spine-cli/tests/cli.rs" }`, **spec 049 included** — that line is in 049's own frontmatter at line 22. The review's own phrasing concedes the condition ("*If* `tests/cli.rs`'s owning spec is 049 or 037"), and it is not: neither establishes it.

The spec 052 consequence does not follow either. `couple` names the crossing from the **owner set of the changed path**, and spec 118 is in that set by its own edge. The gate reports `OK: 4 path(s) checked, no drift`, and CI's `self-governance` job runs the same verb.

### Spec 049's `spec.md` not updated; missing a back-reference — FALSE POSITIVE

There is no such key. `extended_by` and `refined_by` do not exist in the frontmatter grammar: `grep -rn 'extended_by\|refined_by' crates/spec-spine-types/src/` returns nothing, and the known-keys list in `frontmatter.rs` carries `amends`, `amends_sections`, `amends_verification` and `amendment_record` and no back-reference of any kind. A spec cannot be asked to satisfy a field the compiler would reject with exit 3.

The mechanism is the one `CLAUDE.md` states under "How a file gets claimed": an `extends`-carried unit is a first-class claim, the unit does not have to appear in the target spec's territory, and the target spec is not edited. That is the whole point of extending rather than amending, and it is why **no** `Spec-Drift-Waiver:` is declared here or needed. `spec-spine verify 049` still passes on this branch (9 commands), which is the direct check that 049's own acceptance is untouched.

Two of three findings in this pass are about authority mechanics that the corpus answers mechanically, and both were resolvable by running a verb rather than by reading the diff.

bartekus:
Third pass. No new code path is named, and both spec-compliance items are the same two already answered with measured output, now explicitly hedged as unverifiable from the review's own context ("depends on spec 053's content (not shown in REPO CONTEXT)", "REPO CONTEXT does not show per-spec ownership tables, so I cannot confirm base ownership"). Both are answerable by running a verb, and were:

- **Spec 053** is `L-007` and requires a `depends_on` entry to name a **lower** ordinal than the declaring spec. It contains no rule about gaps, and no rule anywhere requires contiguous ordinals. 118's entries are 035, 037, 049, all backward. `lint --fail-on-warn` exits 0, and the knob is off by default in `spec-spine.toml` besides. The gap is the reserved 115 to 117 identities on the wave branch at `487bbd9`.
- **`tests/cli.rs` ownership**: `spec-spine index owner crates/spec-spine-cli/tests/cli.rs` reports `001-compile-registry  floor  crates/spec-spine-cli (package manifest)` and no establisher at all. Spec 049 carries this same unit crediting 001, in its own frontmatter at line 22. So the review's conditional resolves to valid: spec 001 does own it, as the floor.

### The one new item: `let _ = pump.join()` discards a panic

Graded in the review itself as "a design note, not a bug", and I agree with both halves of its reasoning: `JoinHandle::join` returns `Err` only on a panic, the thread body is `io::copy` over a `ChildStdout` and a `Stderr`, and `io::copy` reports every failure it can encounter as `Err` rather than by panicking. What is discarded is therefore a value that cannot be produced by this thread body short of an allocation failure.

Left as it is deliberately, because there is no defensible thing to do with it. Spec 118 D-3 decides that a failed forward is not a failure of the verb: the verdict is computed from exit statuses, and failing an acceptance run because its log could not be written would report a defect in the corpus that is not there. A panic in the forwarder is the same class of event one step further along, so promoting it to a verdict change would contradict D-3, and promoting it to a diagnostic on stderr would be writing to the channel that had just failed.

Both remediation rounds are spent and both were spent on accepted findings: `cbc92e1` (pump not joined on the `wait()` error path, pass 1) and `00ba7bb` (unreachable `if let Some` on the stderr handle, pass 2). Every finding across all three passes is accounted for above.

PR #261, opened 2026-09-19
fix(118): drain both child pipes when the log destination fails

bartekus:
## What

Spec 118 D-3: an inability to deliver an acceptance command's logs must not change the acceptance verdict. The merged implementation (#260, `71a423a`) read "abandon the forward" as "stop reading", which is a different promise. At `71a423a` an undeliverable log decides the verdict by three routes. This corrects the implementation; **D-3 is not weakened**, it is implemented.

## Reproduced at `71a423a`

A consumer that reads the opening `[verify] $ ...` transcript line and then closes the parent's stderr:

| fixture | consumer open | consumer closed at `71a423a` |
| --- | --- | --- |
| `awk` writing ~1.3 MB to the child's **stderr** | ~0.5 s, exit 0, valid envelope | **never completes** (killed at a 30 s deadline) |
| `sleep 0.2; true` | 0.21 s, exit 0, valid envelope | **exit 101**, stdout empty |
| ~1 MB on the child's **stdout** | exit 0, valid envelope | **exit 101** |

Mechanisms, all three visible in `cmd_verify.rs`:

1. `io::copy` returns on the first failed destination write. `child_err` stayed open while `child.wait()` waited, so a child with more to say than a pipe buffer blocked on its own write forever and the parent blocked on it.
2. The `--json` transcript went through `eprintln!`, which unwraps its write. The `[verify] exit 0` line panicked before the envelope was ever produced.
3. The forwarding thread owned its end of the child's stdout pipe and dropped it on the failed write, handing the child an `EPIPE`. That does not merely lose a log; it changes the child's own outcome, which is precisely what D-3 forbids.

## Correction

`drain_to_stderr` replaces `io::copy` on both streams: read to EOF through a fixed 16 KiB buffer whatever the destination does, discard what cannot be delivered, and never close a child pipe because forwarding failed. The `--json` transcript goes through a non-panicking stderr write.

Preserved and asserted: bounded memory (fixed buffer, nothing accumulated), concurrent draining (one thread per stream), normal diagnostic delivery, one JSON envelope on stdout, execution order, stop-on-first-failure, and the whole of the non-`--json` mode. Local to `verify`; no process framework, no DTO or schema change, the `core` seam untouched.

## The discarded `pump.join()`, reassessed

A **destination write failure**, a **forwarding thread panic**, and an **unreadable child pipe** are three different facts. The old `let _ = pump.join()` made them one, and a panic is not evidence that stderr stopped accepting bytes. `Drained` now keeps them distinct. D-3 covers the destination failure alone (and is silent by construction: there is nowhere left to report it). The other two get a best-effort stderr warning naming the stream and the condition.

**Open, and deliberately not decided here (D-5):** whether a forwarding panic or an unreadable child pipe should affect the acceptance verdict. Today none of the three does, which preserves the shipped behaviour and D-3's reasoning for the case D-3 actually decided. Deciding the other two is a policy change, not a correction, and is not needed to fix what was measured.

## Tests

Four closed-consumer regressions plus an open-consumer control in `tests/verify_streams.rs`, each bounding its wait, asserting the process status, and parsing the whole of stdout as one envelope. The consumer closes **after** the first transcript line, so what is exercised is forwarding and completion rather than start-up. On a timeout the harness kills and reaps the parent (which closes the pipes its own child is blocked on), and joins both reader threads before asserting, so nothing is left hung; confirmed by `pgrep` after the fail-first run.

Fail-first at `71a423a`, with only the source reverted:

```
a_closed_stderr_consumer_does_not_change_a_quiet_passs_verdict ... FAILED  (left: Some(101))
a_closed_stderr_consumer_does_not_hang_a_child_flooding_its_stdout ... FAILED  (left: Some(101))
a_closed_stderr_consumer_does_not_hang_a_child_flooding_its_stderr ... FAILED  (did not finish within 30s)
a_closed_stderr_consumer_keeps_a_failing_commands_details ... FAILED  (did not finish within 30s)
an_open_stderr_consumer_receives_every_forwarded_byte ... ok
test result: FAILED. 7 passed; 4 failed
```

Spec 118's `## Verification` gains three end-to-end lines (quiet + closed, volume + closed, and the open-consumer control). The first two are red at `71a423a` (AssertionError `101`, AssertionError `-9`); the control is green there on purpose.

## Spec

Draft / complete, unratified. §3.2 gains the drain obligation, §3.5 a closed-consumer test item, and **D-4** (the measurement and what D-3 does and does not excuse) and **D-5** (the three drain outcomes, and the open policy question) join the decision history. No requirement is withdrawn.

## Evidence

- Gate green before the commit: `compile`, `index`, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (3 paths, no drift), `cargo test --workspace --locked`, `cargo clippy ... -D warnings`, `cargo fmt --all --check`.
- `spec-spine verify 118-...` exit 0, 12 commands.
- Affected acceptance, all exit 0 with stdout/stderr retained: 049, 052, 103, 035, 037, 105, **106**.
- The earlier unexplained `verify 106` failure did **not** reproduce on this run (exit 0). It was run once; no cause is claimed and none is ruled out.


bartekus:
Read all feedback pages: 0 inline comments, 0 submitted reviews, 1 issue comment (the automated review). Accounting for each of its three notes, all raised as design notes with "no findings requiring action":

1. **`Ok(drain_to_stderr(...))` is type-asymmetric, so `note_drain`'s `Err` arm is unreachable for stderr.** Correct, and addressed in 7e16db1 with a comment: the unreachability is a property of *where* the drain runs (a panic there unwinds `run_one` rather than being caught), not a claim that the stream cannot fail, and `InputFailed` still reaches `note_drain` from both streams. Comment only, no behaviour change. Worth taking because D-5 is precisely about not conflating these conditions.
2. **The stderr lock is re-acquired per 16 KiB chunk.** Deliberate, kept. Holding it across a blocking `src.read` would starve every other stderr writer for as long as the child is silent, and §3.2 promises no ordering between the two streams in any case.
3. **The test harness's 10 ms poll.** Test-only and bounded; kept.

Gate re-run to completion on 7e16db1: compile, index, check --fail-on-unresolved --fail-on-warn, lint --fail-on-warn, index coverage --fail-on-untraced, couple, cargo test --workspace --locked, clippy -D warnings, fmt --check, all exit 0. `spec-spine verify 118-...` exit 0, 12 commands.

bartekus:
Round 2 of the automated review raised one supported finding, and it is a real one about this PR's own reasoning: the comment before `child.wait()` claimed both drains had reached EOF. Only the stderr drain has; the stdout drain runs on the pump thread and is joined *after* the wait. The code was correct and the justification was stated wrongly, which matters here because the ordering is the thing this PR is about. Fixed in 45d99e1: what prevents a full pipe at the wait is that both streams are being read concurrently with the child, not that either has finished. Comment only, no behaviour change.

Its other sections raised no action: the `command` interpolation into warning text is a stderr diagnostic with no injection surface, the coupling discipline is satisfied, `DRAIN_BUF` meets §3.2's memory bound, and the 16 KiB lock granularity is acceptable given §3.2's explicit no-ordering promise between the two streams.

Gate re-run to completion on 45d99e1: check --fail-on-unresolved --fail-on-warn, lint --fail-on-warn, index coverage --fail-on-untraced, couple, cargo test --workspace --locked, clippy -D warnings, fmt --check, all exit 0. `spec-spine verify 118-...` exit 0.

bartekus:
Round 3 of the automated review raised two supported notes; both are fixed in aa341eb.

1. **The control test's byte margin was thin.** It asserted `stderr.len() > 2_000_000` against 2 x 1_001_000 minus the transcript line the harness had already consumed: about 1800 bytes of headroom, enough that a longer fixture command would turn a correct run red. It now counts the forwarded lines exactly (2000 lines of 1000 `X`), which is the property it meant to assert and carries no margin. The corresponding acceptance line had the same shape against one stream and got the same treatment.
2. **D-4 and D-5 were appended after D-2**, so the ledger read D-1, D-2, D-4, D-5, D-3. Reordered; no text changed.

Its other sections confirmed rather than flagged: the discard path acquires no lock and writes no bytes once `Drained::Discarded` is reached; drain -> wait -> join is correct because the pump reads concurrently throughout; the `Err` arm unreachable for stderr is documented; the 16 KiB lock granularity is a design note against a channel with no throughput commitment, and is kept.

Gate re-run to completion on aa341eb, all exit 0, and `spec-spine verify 118-...` exit 0. Both derived shards regenerated and committed with the change.

bartekus:
Round 4, both notes accounted for.

1. **`InputFailed` shadows `Discarded` when one stream meets both.** A real gap, and against D-5's own claim rather than against the code: the behaviour is right (the read failure is the one whose consequence is unknown and the one D-3 does not excuse, so it is the fact worth reporting), but D-5 said the three outcomes are kept distinct and this is the one place they collapse to two. Stated explicitly in 1563df5, in the variant's doc comment and in D-5, as a deliberate precedence. No behaviour change.
2. **A second comment at the `note_drain("stderr", ...)` call site.** Declined: the comment three lines above, added in 7e16db1, already says the `Err` arm is structurally unreachable there and why. A restatement at the call site would be the same fact twice, and `debug_assert!(err_drained.is_ok())` would assert something the type-level construction already guarantees, which is the shape this corpus calls a vacuous assertion.

Its remaining sections confirmed rather than flagged: coupling compliant with no waiver needed or missing, no security surface, and the 16 KiB lock granularity negligible per spawn.

Gate re-run to completion on 1563df5, all exit 0; `spec-spine verify 118-...` exit 0; both derived shards regenerated and committed.

This is the last remediation round. Successive review passes are now returning documentation-precision notes on text this PR added rather than new code paths, so I will merge once checks settle green on 1563df5.

PR #263, opened 2026-09-19
fix(118): the outer supervisor owns the fixture's cleanup

bartekus:
Corrective PR on top of #262. Test infrastructure and the acceptance commands
that mirror it; the verb, the verdict semantics, the existing stream
regressions, the open-consumer control and D-5's explicit deferral are
untouched.

## The two cleanup failures

**The Rust outer bound reported a failure it did not clean up.** `within` ran
the fixture on a worker thread and panicked on the *waiting* thread when the
bound expired. The worker still owned the `Fixture`, still blocked, so no `Drop`
guard ran; and the fixture is deliberately in a process group of its own, which
exiting the test binary does not reach either.

Measured, with the group signal injected out and a ten-minute inner budget under
a two-second outer bound: the case reports at 7.02 s and the fixture leader is
still alive afterwards, reparented to init in its own group
(`ps`: `66493 1 66493 sleep 600`).

**The Python watchdog's guard did the opposite of the invariant it cited.**
`p.poll() is None and os.killpg(...)`: `poll()` *reaps*. If the leader exits
while a descendant still holds the pipes, the guard reaps the leader, concludes
the tree is finished, and skips the group signal.

Measured on an isolated fixture whose descendant sleeps two seconds: a 0.2 s
watchdog returned after **2.01 s** with the descendant's delayed side effect
**present**. Corrected: **0.21 s**, absent.

## Ownership and shutdown

Termination authority belongs to the outer supervisor, never to the worker whose
deadline is under test.

- A shared `Tree` handle is created by the supervisor *before* the worker starts.
  The worker publishes the leader's pid at spawn and reaps only inside that
  lock, publishing `Reaped(status)` in the same critical section as the `wait`
  that produced it.
- A signal can therefore never follow a reap, so the negative pgid cannot come
  to name a recycled pid. That is D-6's invariant, now enforced by a lock rather
  than by a `poll()` that breaks it.
- On overrun the supervisor signals the group itself, then allows a bounded
  grace for the worker to return (the group signal closes the pipes it is
  blocked on) so the leader is reaped by its owner. Whether that happened is
  reported, not assumed.
- Python mirrors it: one lock over signalling and reaping, a bounded `wait` that
  releases between attempts so the watchdog can always fire, and no
  `communicate()` (it reaps outside the lock).

Consolidated rather than layered: three inline watchdogs become one claimed
script, `scripts/verify-closed-consumer.py`.

## New bounded regressions

- `a_broken_inner_deadline_is_terminated_by_the_outer_supervisor` — a
  deliberately broken ten-minute inner budget under a two-second outer bound.
  Asserts the supervisor signalled the tree itself, the leader was reaped and
  killed, and the descendant's delayed side effect is absent when checked past
  the moment it was scheduled for.
- `an_exited_leaders_descendant_on_the_pipes_is_still_terminated` — a leader
  that exits while a descendant retains the pipes. Green on `c7e6a75`: what was
  defective there was D-6's prose, not the harness, and this is what stops the
  description drifting again.
- Two matching acceptance lines in `## Verification`.

## D-6 corrected

D-6 said an exited leader with surviving descendants "is not cleaned up by this,
because by then the leader is reaped". That contradicts the harness the same
decision specifies: readers are awaited *before* the leader is polled, so at the
moment a reader overruns the leader is unreaped and the group signal is safe.
D-7 replaces that paragraph, records what the Rust harness and the Python
watchdog each guarantee **separately**, and removes the unmet cleanup obligation
rather than restating it as an accepted limitation.

## A flaky symptom, recorded and not papered over

`verify_streams` failed five cases with `NoTranscript` twice during this work and
passed in between. No deadline was raised. What the evidence shows:

- Once the **test binary itself** sat at 0.0% CPU with **no child processes at
  all** for ~4.5 minutes, then ran all fifteen cases in **8.01 s** and passed.
- The other time, the five failures were exactly the cases that exec
  `spec-spine` *under a deadline*; every case that execs the same binary with no
  deadline passed in the same run by waiting it out, and the `/bin/sh` safeguard
  cases were unaffected. The enriched message says the leader was **still
  running** in all five. The same binary answered `--version` in 0.00 s
  immediately afterwards and had not been relinked since 01:56.
- Host: six to eight concurrent `cargo`/`rustc` processes from unrelated work,
  load average 6-12, `syspolicyd` at 35-52% CPU.

**The cause is not established** and none is claimed. `DEADLINE` stays at 30 s;
`NoTranscript` now carries the leader's status at termination, which is what
told the two recurrences apart.

## Gate

`lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` all exit 0
(`couple`: 3 paths checked, no drift). `cargo test --workspace`, `clippy -D warnings`,
`cargo fmt --check` all exit 0. `spec-spine verify 118-...` exits 0, passed
(14 commands).

118 stays `draft` / `implementation: complete`. Not ratified here.


bartekus:
Reviewed both findings against the source.

**Finding 1 (`looks_killed_by_the_harness` is a stub; the `NoTranscript` else-branch is dead code): declined, false positive.**

There are two `cfg`-gated definitions. The review quotes the `#[cfg(not(unix))]` fallback; the one that compiles here is:

```rust
#[cfg(unix)]
fn looks_killed_by_the_harness(status: &std::process::ExitStatus) -> bool {
    use std::os::unix::process::ExitStatusExt;
    status.signal() == Some(9)
}
```

which is exactly the predicate the review recommends. The `true` body is `#[cfg(not(unix))]`, and every case that consumes this predicate is `#[cfg(unix)]`.

The suite also disproves the dead-code claim empirically. `stuck()` returns `ReaderStuck` **only** when the predicate is false:

```rust
if looks_killed_by_the_harness(&status) { NotFinished { .. } } else { ReaderStuck { .. } }
```

and `an_exited_leaders_descendant_on_the_pipes_is_still_terminated` asserts `ReaderStuck` — a leader that exited 0 on its own while a descendant held its pipes. That test passes locally and in `build · test · clippy`, so the else-branch is live. The other branch was observed too: a local `NoTranscript` failure printed "the leader was still running when the harness terminated it".

**Finding 2 (`--transcript-prefix ""` slips past the exclusivity guard): confirmed and fixed.**

`bool("")` is `False`, so an explicit empty prefix read as an absent flag, and `first.startswith(b"")` would then have asserted nothing. The guard now tests presence rather than truthiness, and an empty prefix is rejected outright rather than accepted as a vacuous assertion:

```python
if (args.transcript_prefix is not None) == args.no_transcript:
    ap.error("exactly one of --transcript-prefix and --no-transcript")
if args.transcript_prefix == "":
    ap.error("--transcript-prefix must not be empty: every line starts with it")
```

Probed: `--transcript-prefix ''`, neither flag, and both flags are each rejected; the valid shape still runs.

The three design notes are accurate and no change is intended for them.

bartekus:
Second review round, evaluated against the source.

**1. `kill_group()` returns `True` on `PermissionError`: confirmed and fixed.**

The two exceptions are not the same answer and were being collapsed. `ProcessLookupError` means nothing was left in the group to signal, so the postcondition already holds; `PermissionError` means the signal was *refused*, and returning `True` there claims a termination that did not happen. They are now separate arms. Probed directly: `PermissionError -> False`, `ProcessLookupError -> True`.

**2. Mutex held across `child.wait()`: accepted as-is, now stated in the code.**

The review's own conclusion is the right one, and the reason is now written where the tradeoff lives rather than left to be rediscovered:

> The lock is deliberately held across `wait`, which blocks a concurrent `Tree::kill_group` for the duration. That window is the price of the invariant, and it is bounded: the `wait` is preceded by a `SIGKILL`, which cannot be caught or blocked, so the leader is already dying. Releasing the lock to wait and re-taking it to publish `Reaped` would narrow the window by reopening the signal-after-reap race this type exists to close.

**3. `assert_eq!(stream, "stdout")` may be order-dependent: not a race, and now documented.**

The descendant does inherit both pipes, but this case runs with the consumer **closed**, and `try_run_fixture` drops the stderr reader on that path rather than draining it:

```rust
let err_rx = if close { drop(err); None } else { Some(drain(err)) };
...
let stderr = match err_rx { None => Vec::new(), Some(rx) => ... };
```

With `err_rx` as `None` the stderr arm cannot reach `stuck` at all, so stdout is the only stream the harness is ever waiting on here and the only one the failure can name. A comment now says so at the assertion.

**Minor: `reap(timeout=None)` in the happy path.** Confirmed correct, and the callsite now records why: `reader.join()` returns only on end-of-file on the child's stdout, which needs every member of the tree to have released that pipe, so by the time `reap` runs the leader has either exited or been killed by the watchdog.

**Minor: non-Unix `looks_killed_by_the_harness`.** Left as is. Off Unix there is no `ExitStatus::signal()` to consult, so the distinction is genuinely unavailable; the existing doc comment already records that the label is a diagnostic with known ambiguity, and every case that consumes it is `#[cfg(unix)]`.

Gate re-run to completion locally before this push.

bartekus:
Third review pass. Both remediation rounds for this PR are spent, so this is an assessment rather than another push; one finding is left **open** and named here rather than quietly accepted.

**1. `Tree::kill_group` leaves the state `Live` after signalling: not a defect, and deliberate.**

`Live` means *spawned and unreaped*, which is exactly true after a `SIGKILL` that nobody has yet `wait`ed on. Moving off `Live` at the signal would be the bug: the reap has not happened, and the state is what licenses a later reap to publish `Reaped`.

The `signalled` race described is not reachable on this path. `supervise` signals *before* it waits:

```rust
let signalled = tree.kill_group();
let _ = rx.recv_timeout(OUTER_GRACE);
```

The worker owns the only `Fixture`, and it is still blocked when `kill_group` runs, so its `Drop` cannot have published `Reaped` first. The review reaches the same conclusion.

**2. Python `ProcessLookupError` returning `True`: answering a contract that was already changed.**

The return value is no longer "a signal was sent". This PR redefined it, in the docstring the review quotes: *"True iff the tree is now terminated."* Under that contract `ProcessLookupError` is correctly `True` (nothing was left in the group, so the postcondition holds) and `PermissionError` is correctly `False` (refused). No caller needs to distinguish "I sent it" from "it was already gone", because neither changes what the caller does next.

**3. `reaped.expect()` is load-sensitive: CONFIRMED as a risk, and left OPEN.**

This one is right, and I am not going to argue it away. `OUTER_GRACE` is a fixed 5 s, and this PR itself documents a host that stalled a binary for ~4.5 minutes. On such a host the worker could fail to return inside the grace, `reaped` would be `None`, and `expect` would panic — a noisy false failure in a test whose subject is not the thing that failed.

Two facts bound the blast radius, and neither makes it go away:

- The tree is signalled **before** the grace, so the cleanup the case exists to guarantee has already happened when the grace expires. `signalled` and the delayed-side-effect witness are both asserted independently of `reaped`.
- So the exposure is a flaky assertion about *who reaped*, not a missed termination or a silent pass.

It has not been observed: the case passed locally across repeated runs and in `build · test · clippy` on three separate head shas. That is why it does not meet the "confirmed correctness defect" bar that would stop this merge — but it is a real weakness in a test added by this PR, and it is being recorded as open work rather than folded into the spec as an accepted limitation. The likely remedy is to treat `reaped` as diagnostic (reporting it, asserting only `signalled` and the witness) instead of `expect`ing it, which is a change to an assertion rather than to a guarantee.

PR #265, opened 2026-09-19
chore(118): ratify (draft -> approved)

bartekus:
Ratifies spec 118, `The verdict is the only thing on stdout`, on human approval of its current contract.

## What this changes

One authored line: `status: draft` -> `status: approved` in `specs/118-the-verdict-is-the-only-thing-on-stdout/spec.md`. `implementation` stays `complete`. The two regenerated shards for 118 (registry and index) ride along because the status is compiled into them.

No requirement, decision entry, acceptance command, or implementation is touched.

## What is approved

The contract as built, **including the deferral D-5 states explicitly**: whether a forwarding panic or an unreadable child pipe should affect the acceptance verdict is a policy change rather than a correction, it was not needed to fix what 1.1 and D-4 measured, and it is left to a later spec. Today none of the three outcomes changes the verdict.

## Evidence

The build and its four corrections merged through #260, #261, #262, #263 and #264, the last as `0fe2e89769d571d36fee1ecf080b766edf5045bd`. Post-merge `spec-spine verify 118-the-verdict-is-the-only-thing-on-stdout` passed once: 14 commands.

Gate on this branch, all exit 0: `compile`, `index`, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple --base origin/main --head HEAD` (`OK: 1 path(s) checked, no drift`), then `cargo test --workspace --locked` (including the 19 `verify_streams` regressions and 96 `cli` tests), `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all --check`.

## Known open item, carried not resolved

#263's third review pass recorded one finding as OPEN rather than accepted: `reaped.expect()` in the outer-supervisor case is load-sensitive against a fixed 5 s `OUTER_GRACE`, on a host this spec itself documents as having stalled a binary for minutes. It has never been observed failing, the termination the case exists to guarantee is asserted independently of `reaped`, and the exposure is a flaky assertion about who reaped rather than a missed cleanup. It is not a reason to withhold ratification and is not addressed here.

No waiver is declared or needed.


bartekus:
Graded against source: the one finding is a **false positive**, and the review states its own precondition as unverified ("cannot be confirmed as a hard violation without reading those specs' acceptance criteria").

**No spec in this corpus carries a co-filed acceptance artifact, because there is no such artifact.**

```
$ find specs -mindepth 2 -maxdepth 2 -type f ! -name spec.md | wc -l
0
```

All 116 spec directories contain exactly `spec.md` and nothing else. The review's suggested check ("every other approved spec in the repository should be inspected for whether it carries an accompanying acceptance record") returns the same answer for all 115 already-approved specs as it does for 118.

The two specs named hold the opposite of what is inferred:

- **049 (verify-declared-acceptance)** puts acceptance *inside* `spec.md`, as the `## Verification` block `spec-spine verify <id>` executes. 118 has one, and it passed post-merge on `0fe2e89`: 14 commands, exit 0.
- **042 / 023 (attestation)** emit a *derived* document to `.derived/attestation/attestation.json`, which is gitignored (`.gitignore:30`). It is never committed, for any spec, so its absence from a diff is the expected state and not a missing co-artifact.

Nothing in either spec conditions a `status` transition on a co-filed document. The lifecycle flip is validated by the corpus's own verbs, and `self-governance (compile · index · lint · couple)` is green on `3e92bc7` along with every other required check.

Merging.

PR #266, opened 2026-09-19
feat(116): shepherd reads every reviewer, on every path to merge

bartekus:
## Summary

`/shepherd` read one of the three places a GitHub reviewer can write. Step 3b
fetched `pulls/<n>/comments` (line-anchored comments only), while this
repository's AI review pass posts with `gh pr comment` to `issues/<n>/comments`,
and a submitted review's body goes to `pulls/<n>/reviews`. Step 1 also routed an
all-green PR straight to the merge checkpoint, so the one path where a
reviewer's comment is the only thing between a branch and the default branch was
the path that never looked, and the report then claimed `Review threads: none`.

Step 3b now reads all three endpoints with `--paginate`, redirecting each to a
file outside the worktree so the read's own exit status is what gets checked
rather than a downstream `--jq`. A failed read gets one retry and then stops the
run before Step 4, naming the endpoint and preserving the diagnostics; a partial
read never establishes complete coverage. Step 1's green route passes through
Step 3b, and Step 4 states the read as a precondition. The report carries
`could not read <endpoint>` beside spec 082's unchanged
`not read: stopped at CRITICAL`, with successful coverage on its own
`Thread reads:` line, and `none` now means all three reads succeeded and found
nothing.

Spec 082's budget is unchanged at two rounds. Reading, classifying, retrying and
rejecting a false positive cost none; a confirmed authorized fix inside the
territory costs one even when CI was green; an exhausted budget with substantive
findings open is a stop, not a clearance. Decisions D-7 through D-11 record all
of this in the spec before the code.

The edit is made once in `kit/.claude/skills/shepherd/SKILL.md` and propagated by
`scripts/gen-agent-trees.py` and `scripts/gen-kit-embedded.py`; no generated copy
was hand-edited. Editing `kit/**` is a hashed input, which is why every index
shard is regenerated here. No helper executable and no follow-up spec: the
residue is written down as the limitation in section 4, that a test over shipped
instructions cannot prove a session obeyed them.

## Testing

Five new `shepherd_*` tests in `kit_skills.rs` assert the per-endpoint
pagination (not a count of `--paginate`), the positive green-path routing through
Step 3b, the failed-read stop, the three distinct report values, and the
preserved two-round budget. All five were verified red against the parent's skill
text before the edit, each naming the line it read.

The gate from `AGENTS.md` was run before every commit and is green:
`lint --fail-on-warn`, `index coverage --fail-on-untraced`,
`couple --base origin/main --head HEAD` (6 paths checked, no drift), then
`cargo test --workspace --locked`, `cargo clippy --workspace --all-targets
--locked -- -D warnings`, `cargo fmt --all --check`, all exit 0.

`spec-spine verify 116-shepherd-reads-every-reviewer`: passed, 35 commands,
exit 0.

Spec 116 is `status: draft`, `implementation: complete`. The ratify flip is a
human's.


bartekus:
## Shepherd dispositions, head `b26de68`

Read on three endpoints, `--paginate --slurp`, each command's own exit status checked before its output was parsed: `pulls/266/comments` ok (1 page, 0 items), `issues/266/comments` ok (1 page, 4 items), `pulls/266/reviews` ok (1 page, 0 items). `reviewDecision` empty.

### Confirmed and fixed in `b26de68`

- **`--slurp` not asserted in the Rust suite** (04:17 pass). Correct. `shepherd_reads_all_three_endpoints_paginated` is now `…_paginated_and_slurped` and requires both flags on *each endpoint's own command line*; a file-wide count would have been green for a flag sitting in the surrounding prose.
- **The `--jq` explanation was itself wrong** (self-identified, not raised by a reviewer). `--jq` is an option of `gh api`, parsed by `gh` and run in the same process; the status the shell sees is still the read's own, so it does not hide anything. The ban on `--jq` pinned a false reason and witnessed nothing about coverage. Removed from the skill, from `kit_skills.rs` and from the acceptance block, and replaced with the obligations that do keep a thread unread: pagination, a documented output shape, a parse before the document is believed, every item of every page inspected, and a count of feedback items rather than of outer pages. Recorded as D-13; D-5 and D-12 keep their text with the superseded sentences marked in place.

### Rejected, against source

- **"`--slurp` flattens to one array, so `[.[][]]` corrupts the list"** (04:05 pass). False, and measured again on this PR: `jq 'length'` on the slurped `issues/266/comments` file is `1` (pages) and `jq '[.[][]] | length'` is `4` (items). The top level is an outer array of pages. `[.[][]]` is the flatten; `.[] | .id` is the mistake. `gh` 2.73.0 documents the same: "Pass `--slurp` to wrap all pages of JSON arrays or objects into an outer JSON array."
- **"`--paginate` without `--slurp` produces output no parser can read"** (22:24 pass). Overstated, and D-12 said it too; corrected in D-13. `[{"id":1}][{"id":2}]` is a *stream* of top-level JSON values: `jq` reads it as two inputs and exits 0, `jq -s` combines them; only a single-document parser rejects it (`python3 -c 'json.load(…)'` exits 1 with `Extra data: line 1 column 11`). `--slurp` is required for a **defined** shape, not for an unreadable file.
- **"`at most two rounds` may be absent from the unchanged skill body"** (04:17 pass). Present, 2 occurrences in each of the three copies.
- **"`skill_section(&body, \"Step 3:\")` may panic on a heading without a colon"** (all three passes). The heading is `## Step 3: remediate (at most two rounds, worst first)`. The substring-vs-line-anchored variant of the same note is acknowledged as latent and does not fire; a section-helper redesign is out of this spec's scope (§4).
- **`status: draft` + `implementation: complete` is malformed** (all four passes). This is the shipped pre-ratify state the corpus is built on: `AGENTS.md` step 6 ships with `implementation: complete` and flips `status` in a separate ratify PR. `compile`, `lint --fail-on-warn` and `check --fail-on-unresolved --fail-on-warn` all pass on this tree, and specs 044/045 govern the pair rather than forbidding it.

### Noted, not a defect

- **`Thread reads:` has no per-slot failure token.** §3.3 puts the failure value on the thread line (`could not read <endpoint>`) and requires successful coverage to be reported "separately and in its own words" beside it, which is what the second line is; the Report section's prose says exactly that. A `| err` slot would be a refinement of §3.3, which is a design question for a human rather than a session edit.

### Validation added

The documented reading approach is now exercised by four acceptance fixtures rather than only asserted in prose: `[[]]` is one page and **zero** feedback items, `[[{…}]]` is one and one, `[[a,b],[c]]` is **two** pages and **three** items, and a truncated document makes `jq` exit non-zero and yields no count at all, which is why a parse failure is an unread endpoint and not an empty one.

Gate green locally before the push (compile, index, check, lint, coverage, couple, `cargo test --workspace`, clippy, fmt) and `spec-spine verify 116` passes, 51 commands. All nine required checks pass on `b26de68`.

PR #267, opened 2026-09-20
chore(116): ratify (draft -> approved)

bartekus:
## What this is

The ratify half of spec 116. **#266 delivered the implementation** and merged as `4078694`; this PR records the owner's explicit approval of that spec as implemented, by moving `status` from `draft` to `approved`.

Per `AGENTS.md` step 6, shipping sets `implementation: complete` and a separate PR flips `status`. That flip is the only source change here.

## The diff

| File | Change |
|---|---|
| `specs/116-shepherd-reads-every-reviewer/spec.md` | one line: `status: draft` -> `status: approved` |
| `.derived/spec-registry/by-spec/116-….json` | regenerated: `status` and `shardHash` |
| `.derived/codebase-index/by-spec/116-….json` | regenerated |

Unchanged on purpose: `implementation: complete`, every requirement in §3, the decision history D-1 through D-13 (including the sentences D-13 marks as superseded in D-5 and D-12, which are preserved in place rather than rewritten), the `## Verification` block, the shipped `/shepherd` instructions in all three skill trees and `kit_embedded.rs`, and the tests in `kit_skills.rs`.

## Validation on this branch

Full `AGENTS.md` gate, before the commit: `compile` 0, `index` 0, `check --fail-on-unresolved --fail-on-warn` 0, `lint --fail-on-warn` 0, `index coverage --fail-on-untraced` 0 (96/96 claimed), `couple` 0, `cargo test --workspace --locked` 0, `cargo clippy --workspace --all-targets --locked -- -D warnings` 0, `cargo fmt --all --check` 0.

`spec-spine verify 116` passes on this branch: exit 0, 51 commands.

## Scope

This PR does not reopen the implementation. Findings that would change 116's approved behavior are out of scope here and go to the owner with the evidence preserved. No release accompanies this.

PR #268, opened 2026-09-20
feat(119): an unclassified review failure blocks the merge

bartekus:
## What

The AI review job classified a failed reviewer invocation into two outcomes and its fall-through was a pass. On PR #267 the provider returned `Your organization does not have access to Claude. Please login again or contact your administrator.` The job matched that prose against a token allowlist (`authentication_error`, `permission_error`, `unauthorized`, `forbidden` and three more), found no match, set `review_status=api_failure`, exited 0, and `ci-gate` went green with no review.

The correction is not "recognize this one sentence". It is to invert the default: an invocation failure the job cannot positively classify as transient must block, because a pass that cannot say why it passed is a silent green.

## Classification

Exit status alone selects the branch, before any diagnostic text is read:

| | condition | class | outcome |
|---|---|---|---|
| `rc == 0` | review body has a non-whitespace character | 3 | publish, pass |
| `rc == 0` | empty or whitespace-only body | 6 | `::error::`, block |
| `rc != 0` | a refusal signal | 4 | `::error::`, block |
| `rc != 0` | a recognized transient signal | 5 | `::warning::`, notice, pass |
| `rc != 0` | anything else, including empty diagnostics | 6 | `::error::`, block |

Because no pattern crosses the exit-status boundary, a review whose own prose discusses a 403 handler, an expired OAuth token or a billing path is class 3 and cannot be reclassified as a provider failure.

Within a failed invocation a refusal outranks a transient: a 529 that also says "does not have access" is not an outage that happens to mention organizations. Transient signals are recognized only in the contextual form their emitting layer produces, so a bare three-digit number (a line number somewhere), a bare `fetch failed` (which names no failure) and a bare errno with no transport context are each unrecognized and block. A timeout is not a signal at all: exit `124` and exit `137` are ordinary statuses a child returns on its own, and timeout-sounding prose is text. Bare `quota exceeded` reads as a rate limit in one product and an entitlement cap in another, so it is in neither list.

The diagnostic names the observed signal, never a cause. "The provider refused this invocation" is sayable; the logs record what was returned and do not establish why.

## Publication

Classification success, publication success and job success are three results, and the defect is precisely a step that succeeded while the thing it existed to make visible did not happen.

- A review is published only when the invocation exited 0 **and** wrote something. The post step's condition previously carried a comment promising that guard and tested only the exit status; it now tests the body too.
- The transient notice's `|| echo "::warning::could not post API-failure notice comment"` swallow is removed. That notice is the entire difference between a visible skip and a silent green, so a notice that failed to post must fail the step.
- The fork and Dependabot fallbacks are preserved. Their `GITHUB_TOKEN` is read-only by design, and hard-failing there would red-gate every external contribution and every dependency bump for a permission reason unrelated to the review.

## The harness

`crates/spec-spine-core/tests/ai_review_policy.rs` runs the workflow's own `run:` scalars, never a copy of their patterns. A rebuilt classifier asserts a property of its own copy, drifts from the file that runs in CI, and cannot observe exit statuses, step outputs or publication, which is where this defect lives.

It parses the workflow with `serde_yaml`, selects only the steps a case actually reaches, executes each under the shell semantics the runner applies to it (`bash -e` for a step declaring no `shell:`), resolves `${{ }}` and `if:` from fixture context and from the outputs the classifier actually wrote, and models the implied `success()` that the transient notice's condition relies on. It is bounded by design and panics, naming the step and the expression, on anything it does not support: it is not a GitHub Actions emulator and must not become one. Every case is hermetic, with `claude` and `gh` stubs as the only ones on `PATH`, everything inside a per-case temp directory, a non-credential literal for the token, and no provider, GitHub or network call.

28 fixture rows assert the class, `review_status`, the classification step's exit status, which publication step the outputs select and its exit status, the resulting job outcome, and the distinguishing phrase of each class's annotation.

## Regression evidence

Two cases mutate the step in memory and prove the suite refuses on observed behavior rather than on the presence of a pattern:

- rewriting the class 4 branch's terminator from a failing exit to a passing one, leaving every regex byte untouched, is detected;
- reconstructing the pre-119 classifier (`review_status=ok` from exit status alone, no `review_nonempty`) proves the post step's condition independently refuses to publish an empty review. That conjunct is unreachable by any fixture while the classifier's own guard holds, so without this case it was unasserted defense in depth. Recorded as D-16.

Twelve mutation probes were run against the candidate. Eleven were already caught; the twelfth found the gap above, which this PR closes. With the conjunct removed, the new case is the only one that fails.

## Consequence, stated plainly

Under D-9 a refused invocation now fails the job and `ci-gate` fails with it, so while provider access is refused this repository's PRs do not merge on a green gate. That is the cost side of the inversion and it is accepted: a gate that cannot review is a gate that should not pass. The escape is a human decision, not an automatic one. D-9 governs classes 4 and 6 only and withdraws none of the retained exceptions: a structural skip (non-`pull_request` event, draft, fork, Dependabot, oversized diff) still passes with its visible notice, and a recognized transient still passes when its notice published.

This changes no provider account state and does not restore access. Conversely, restoring access would not close the defect: the misclassification is in the code and survives the incident.

## Territory

Establishes `crates/spec-spine-core/tests/ai_review_policy.rs`. References `.github/workflows/ai-pr-review.yml` (exemplar) and `.github/workflows/ci.yml` (context), both non-owning: `references` is the corpus's only non-owning edge, and claiming the workflow would put a `C-001` ratchet on every `CLAUDE_CLI_VERSION` and `DIFF_SIZE_CAP` edit while not in fact defending the classification rule, which a behavioral test does (D-1, section 2.1).

## Verification

`spec-spine verify 119`: 15 of 15 commands exit 0. `cargo test -p spec-spine-core --test ai_review_policy`: 7 passed, 0 failed. Full gate green: `compile`, `index`, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (2 paths checked, no drift), `cargo test --workspace --locked`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all --check`, and `cargo test -p spec-spine-core --no-default-features --locked`.

Spec 119 stays `status: draft`; ratification is a separate PR.


bartekus:
Thanks. Four points raised; one is a real finding and is fixed in `880d944`, two are declined with evidence, one needs no change.

**Fixed: the matrix asserted its size, not its row set.** `assert_eq!(cases.len(), 28)` passes a matrix that duplicated one row and dropped another, which is exactly how a fixture goes missing in an edit. The suite now asserts the row set is `1..=28` and that no row number appears twice. Fail-first evidence: duplicating row 1 over row 28 fires `every row of 3.8.1 is present exactly once` and no other case notices.

**Declined: `"type": "api_error"` is not an implementation excess over spec 3.6.** Section 3.6 lists three recognized provider error types, and `api_error` is the third; the list wraps across three lines and `spec.md:281` is the one carrying it. The implementation matches the spec exactly, so there is nothing here to reconcile. Whether `api_error` *should* be transient is a different question, and a good one: it is the SDK's catch-all, so the argument that it is weaker evidence than `overloaded_error` or `rate_limit_error` has force. But that is a question about the rule, not about conformance to it, and editing what an in-build spec requires so that a review comment is satisfied is not a move this repository permits (`.claude/rules/adversarial-prompt-refusal.md`). If the narrowing is wanted it belongs in a follow-on spec that says why, with the fixture that proves it.

**No change: the `printf '%s\n'` on the network-errno pipeline.** It is deliberate rather than an inconsistency to align away. That branch is the only line-oriented match of the three: it pipes `grep` into `grep` precisely so that the errno and the transport phrase must land on the *same* line, and terminating the last line is what keeps the two greps agreeing on where a line ends. The other two matches are whole-text and have no such requirement. The trailing blank line the note describes matches nothing in either pattern, as the note itself concludes.

**Noted, no action: the spec-spine compliance and security sections.** Both read the edges and the `api_excerpt` sanitisation correctly.

Gate re-run at the new head: `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` (2 paths, no drift), `cargo test --workspace --locked`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all --check`, and `spec-spine verify 119` at 15 of 15.

bartekus:
Second pass. Taken: one. Acknowledged without change: the rest, with reasons.

**Taken (`bd33397`): every diagnostic match now reads a terminated last line.** The new argument holds where the first framing did not. A `$(...)` capture strips the trailing newline, and POSIX does not require grep to process a final line lacking one, so a refusal or transient-type signal sitting alone on an unterminated last line could in principle go unmatched. The refusal and transient-type matches now pass `printf '%s\n'` like the network branch already did. As the review notes, the direction of that failure was fail-safe: an unmatched refusal falls through to class 6 and blocks, so no silent green was reachable through it, and GNU grep on the runner matches an unterminated last line regardless. Robustness, not a defect repair, and worth two characters.

**No change, `pipefail` in the `elif` arm.** Correct as described, and the review agrees. The conditional-context exception is POSIX shell behaviour rather than a bash extension, so a `|| true` would add a guard against nothing.

**No change, `\f` and `\v` in `api_excerpt`.** Real but theoretical, as labelled. Neither character terminates a line for the `GITHUB_OUTPUT` parser, which is what the `tr` exists to protect, and the heredoc rewrite is a larger change than the exposure justifies. Worth revisiting if a diagnostic ever carries one.

**No change, Markdown rendering of `API_EXCERPT` in the notice.** Cosmetic and confined to a comment body, with no execution path. The excerpt is already flattened and truncated.

**No change, row 28's source position.** The review reaches the right conclusion itself: the set-completeness assertion added in `880d944` catches a missing or duplicated row regardless of order. Row 28 sits beside row 9 deliberately, because the two are the halves of one rule and reading them together is the point.

**No change, `./target/release/spec-spine` in the verification block.** That is this repository's convention, not an ordering slip: `AGENTS.md` names the in-tree binary as what the governed loop runs, and every spec's block invokes it the same way.

**No change, the two performance notes.** Both are labelled non-blocking and both are right that they are not warranted at this size.

Gate re-run at the new head, all green, and `spec-spine verify 119` at 15 of 15.

bartekus:
Third pass. One new finding, taken in `5155157`; the rest have been answered above and are unchanged.

**Taken: an unmodelled `GITHUB_ENV` now refuses instead of going unevaluated.** This is the pass's one genuinely new observation and it is right. The harness created `GITHUB_ENV`, passed it to every step, and never read it back. Nothing is wrong today because no step of this job writes one, but a step that started to would have had its value silently dropped while the suite stayed green, which is exactly the failure mode 3.8 rule 5 exists to prevent. Rule 5 requires an unsupported relevant form to panic naming the step, not to default quietly, so the file is now read back after each step and a non-empty one refuses with the step name and the content. Probed by writing `PROBE=1` to `$GITHUB_ENV` in the classification step: the matrix fails with `step "Run AI Review" wrote to GITHUB_ENV, which this harness does not model`.

**Row 28's source position** has now been raised on each of the three passes and the answer is the same: the set-completeness assertion catches a gap or a duplicate regardless of order, as this pass again concludes, and row 28 sits beside row 9 because the two are the halves of one rule and reading them together is the point. The spec table numbers it 28 because it was added later, by D-15, and renumbering a ratified fixture table to match a vector's order would be the wrong direction of fit.

**The performance note** is the same one as the second pass and carries the same answer: acceptable at this size. The observation that per-case `TempDir` plus `AI_REVIEW_TMP` isolation makes the suite safe under `--test-threads` is correct, and is why the scratch root was parameterised in the first place (3.8's prerequisite).

**The security and compliance sections** read the sanitisation, the stub isolation and the edges correctly, with one wording correction worth making for the record: the workflow is not "unclaimed at base, so changing it in the same PR satisfies co-change". It is unclaimed full stop, and `references` is non-owning (spec 034), so the coupling gate ignores it whether or not the spec changes alongside. That is D-1's deliberate choice and section 2.1 states the hole it leaves: what defends the classification rule here is the behavioural test, not the gate.

This is where I stop remediating. Three passes, one real finding each in the first and third, and the second and third both restating a declined note; the suite is anchored on observed exit status, outputs and publication, and the gate is green at the new head.

bartekus:
Fourth pass. One new finding, and it is a good one. Taken in `2301a95`.

**Taken: 3.6's same-line rule has three cases and only two were fixtured.** D-15 closed this rule by adding row 28, on the reading that it has two halves, a transport phrase with no errno and an errno with no transport phrase. That reading was incomplete. The rule 3.6 states is that the two appear in the *same diagnostic line*, so the third case is both present on *different* lines, and nothing covered it. The implementation is correct, for the reason given: the first grep emits only lines carrying an errno and the second searches those. What was missing was anything that would notice if it stopped being correct. Row 29 is `ETIMEDOUT\nfetch failed`, class 6, and the mutant named in the review, two independent full-text greps, now fails on it while leaving rows 9 and 28 green. The matrix asserts its row set is 1 to 29. Recorded as D-17, which says plainly that D-15 got the arity of its own rule wrong.

That is the second time a review pass here has found a fixture that could not fail, and both were the same shape: a rule with more cases than the matrix enumerated. Worth noting in the record rather than just fixing.

**Declined: `please login` subsuming `please log in again`.** The observation is correct, and the review draws the right conclusion about its consequence: no behavioural difference against any fixture, since both route to class 4. Where I differ is the remedy. 3.5 lists both forms deliberately, as refusal *language* an operator would recognise, and the redundancy is in the regex rather than in the rule. Tightening 3.5 to drop one would be editing what an in-build spec requires on a reviewer's suggestion, which is the move this repository does not allow, and doing it for a change with no observable effect is the weakest possible case for it. If the list should be minimal rather than legible, that is a follow-on spec with a reason.

**Compliance section** reads correctly this time, including that the workflow is unclaimed and that `references` carries no coupling obligation.

Gate green at the new head; `spec-spine verify 119` at 15 of 15.

bartekus:
Fifth pass. No finding survives; all three are settled by reading files the review says it could not see. This is where the review has converged, so this is the last remediation round.

**Finding 1, declined: both dev-dependencies already exist.** `crates/spec-spine-core/Cargo.toml:52` declares `serde_yaml = { workspace = true }` and `:53` declares `tempfile = "3"`. The review is explicit that it could not see the file's contents and reasoned from that absence; the conclusive evidence is upstream of the file anyway, since `build · test · clippy` has been green on every head of this PR and `cargo test --workspace --locked` cannot pass with a missing dev-dependency.

**Finding 2, declined: `implementation: complete` is the correct value at PR time.** This is the repository's documented loop, not a mismatch. `AGENTS.md:138`: "`/ship` opens the PR with `implementation: complete` set once that block holds", and the ratify PR afterwards flips `status: draft` to `approved`. Section 3.10's sentence is about the tree the spec was *filed* on, which is the point of stating which ACs must go red before the build; the build has since happened, and `spec-spine verify 119` passes 15 of 15 on this head. The two axes are independent by design: `implementation` tracks whether the work is done and `status` tracks whether a human has ratified it, which is why this PR carries complete and draft together.

**Finding 3, declined: the heredoc form for `api_excerpt`.** The hazard described is real in general and the review states correctly that the current code is safe, because `tr '\n\r\t' '   '` strips every line terminator and `cut -c1-280` bounds the result before the write. What is proposed is insurance against a hypothetical future edit that relaxes the stripping, and the same argument was made about `\f` and `\v` on the second pass with the same answer: the guard is the `tr`, it carries a comment saying so, and rewriting a safe write to document a boundary that the sanitiser already enforces is not worth another round through a policy file this PR exists to stabilise.

Five passes. Two found real gaps and both were the same shape, a rule with more cases than its fixture table enumerated: the post step's body conjunct that nothing could reach (D-16, found before review) and 3.6's same-line rule that turned out to have three cases rather than two (D-17, pass four). Everything else was settled by reading the source. Merging on green.

PR #276, opened 2026-09-21
feat(096): compaction is a verb, not a session

bartekus:
Spec 096 built: `spec-spine compact`, the corpus-compaction verb, plus the two further defects measured while building it.

## Why the spec exists

Spec 095 removed 27 specs and renumbered 93, by hand. That produced five silent defects, every one a rewrite rule that matched more or less than its author meant, and every one green under `compile`, `index`, `lint`, `couple` and `check` while 355 citations pointed at the wrong document. This verb is those rules written once: a plan in, a rewritten corpus out as data, idempotent, refusing every ambiguity the defects came from, and reporting what it changed per form rather than leaving a six-thousand-line diff to read.

## Two more defects, measured during the build

Both were found by diffing the pre-collapse tree against the merged one: for every citation whose ordinal the map moved, whether the same file still carries the old ordinal.

| # | Defect | Measured |
|---|---|---|
| 6 | The prose-citation rule was **line-oriented**, and a citation wraps: the keyword ends a line and the ordinal opens the next behind a `///` or `#` continuation. | 28 citations in 24 files |
| 7 | A citation may name **more than one ordinal** (`spec 017/021`, `specs 022 and 023`) and only the first was read. | 13 more |

They are recorded in §1.1 with the rest of the family, and §3.3 form 2 now carries the two clauses that answer them: one line break with its comment marker is part of a citation, two is a paragraph boundary and is not; and a separator (`/`, `,`, `&`, `and`, `or`) continues a citation onto a further ordinal, each of which is rewritten.

**Those 41 occurrences are still wrong on `main`.** This PR does not repair them: `compact` refuses a plan naming a spec the corpus no longer has, so it cannot clean up after a renumber that already happened. That repair is a separate change and needs `extends` edges to the ~24 owning specs, because a comment-only citation fix trips `C-001` (measured, not assumed).

## What is built

- `crates/spec-spine-core/src/compact.rs`: the plan, the map and the three reference forms as a pure function. Reads the tree, writes nothing.
- `crates/spec-spine-core/tests/compact.rs`: 24 tests, each of the seven defects with a case that fails against the rule that produced it.
- `crates/spec-spine-cli/src/cmd_compact.rs`: reads the plan, writes the files, prints the report. Refuses a dirty tree unless `--force`; `--plan` writes nothing.
- `lib.rs` gains `compact_json`, `main.rs` the subcommand.

Three decisions the build had to make are recorded as D-5 (the foreign-project exclusion is a named list, not a heuristic), D-6 (the scanned file set is extension-bounded) and D-7 (what idempotence asserts, and why a single run holds it by construction: every match is found against the original text and applied once, simultaneously, never key by key).

One departure from the spec's ordering, worth naming: §3.2's ordinal-collision check runs **before** the compile that §3.8 requires. A corpus with two specs on one ordinal fails validation anyway, but as a generic failure, and defect 1 was precisely a reader who could not see which two ids were in contention.

## Evidence

`spec-spine verify 096` passes. `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `cargo fmt --check`, `clippy -D warnings` and the workspace suite are green. Lifecycle flipped to `implementation: complete`; `status` stays `draft` for a separate ratify.


bartekus:
Addressed in e7aa42e3. Reading the four findings in order:

**1. Garbled doc comment: confirmed and real.** The facade was inserted by anchoring on `pub fn attest_json(`, which sits *after* that function's doc block, so three lines documenting the attestation became the head of `compact_json`'s doc and `attest_json` was left with none. Moved, with the blank line. Worth naming the cause: anchoring a patch on a signature rather than on the doc block that owns it.

**2. `removed_paths` versus what `apply` deletes: confirmed and real.** The report now lists the spec **directory**, which is what actually leaves. A spec is its directory, auxiliary files included, and listing the `spec.md` while the writer deletes the folder hides whatever else was in it from the only list a consumer reads. Recorded as D-8, with the test asserting the directory.

**3. Non-atomic apply: correct, and stated rather than changed.** The three phases run in order with no rollback. What makes that recoverable is §3.8's dirty-tree refusal: the tree was clean when the run started, so `git checkout .` restores it exactly. The refusal message now says so, and D-9 records why a transactional writer is not worth building to buy a property `git` already has.

**4. `verify_cli_blocks` exported but not called: correct observation, and the answer is in the spec text you could not see.** §3.5 is conditional: *"Where a rule applies only inside a `verify:cli` block, the block MUST be located by scanning lines."* No rule in this module is block-scoped. Form 3 is keyed on the command that takes a spec id, which is a stronger test than "inside a block" and needs no scoping, so the pre-096 design that defect 4 broke is not reinstated. The locator is `pub` and asserted because §3.5 requires it for the rules that *are* scoped, and spec 097's shell-word form is the first of them. The doc comment now says all of that, so the next reader does not have to ask.

On the note about the new verification filters also matching `a_citation_broken_across_two_blank_lines_is_not_a_citation`: agreed and intended. Both tests must pass, and the filter asserting a non-zero pass count is what stops a filter that matched nothing from reading as a pass.

bartekus:
Second round addressed in cf6055ea. Two of the four were real.

**Renamed spec directory orphaning non-scannable files: real, and now a refusal.** The rewrite moves a spec directory file by file and the file set is extension-bounded, so a binary beside the `spec.md` would have stayed at the old path while the document moved, splitting one spec across two directories. `compact` refuses and names the file (D-10), with a test that puts a `.png` next to a renumbered spec. A removed spec was never affected: its whole directory goes.

**Dead `matched_any`: real smell, and it hid an untested behavior.** Removing it let the behavior be stated instead: each ordinal in a list is decided on its own, so `spec 777/002` rewrites `002` and leaves `777`. That is the intended reading (a list is a list of citations, not one citation whose first member speaks for the rest) and it now has a test covering both a leading unmapped ordinal and a trailing one.

**Four-digit ordinals for ≥1000 survivors: correct about `{:03}`, not reachable.** `refuse_ordinal_collision` demands a distinct three-digit ordinal per spec, and there are exactly 1000 of those, so a corpus that reaches the format string has at most 1000 specs and `i` at most 999. A 1001-spec corpus is refused by pigeonhole before the map is built. The guard is that refusal rather than a clamp; the comment now says so at the line that looks unguarded, since "structurally unprotected" was a fair reading of the code as written.

**`skip_gap` panic: not reachable, and the finding's own wording says why.** A slice at the *leading* byte of a multi-byte character is a valid boundary; the panic would need the index to land on a *continuation* byte (0x80–0xBF). It cannot: every byte the walk steps over is ASCII (space, tab, CR, LF) and the caller enters on a boundary, so the walk always stops on one. The doc comment records the argument.

On the performance notes: `scan_full_ids` being O(bytes × keys) and the `to_ascii_lowercase` copy are both accepted for an offline one-shot verb, and worth having on the record for whenever the corpus is large enough to matter.

PR #277, opened 2026-09-21
feat(097): a path leaves the corpus the way a spec does

bartekus:
Spec 097 built: a path retires the way a spec id does, through the same verb, the same plan and the same report.

## Why it is a second half rather than a second verb

Spec 096 mechanizes one retirement: a spec id leaving the corpus. On 2026-09-20 this repository did the other one by hand, `.claude/rules/` leaving for a section of `AGENTS.md`, and it touched 81 references in 46 files, three frontmatter units, one hashed-input glob, one hook path pattern, six acceptance lines, three Rust tests and 122 derived shards. `compact` reached none of it: its three forms are all spelled `NNN-slug` or three digits, and a path is spelled as a backticked citation, a bare word, a glob, a YAML value, a shell word and a Rust string literal.

So the plan gains a `retire` section rather than the tool gaining a sibling. The two share the plan file, the refusal set, the idempotence requirement and the per-form report; splitting them would mean maintaining two of each.

## What the build had to get right

**A replacement per form, not per path.** The four rule files each became a different section of one document, the directory became a phrase, and the glob became nothing at all. A form with no rule is a refusal, never a silent skip: a form that fires zero times is spec 096's defect 2, and a form never declared is the same defect one step earlier.

**A withdrawn unit is named, never inferred.** There is no grammar in this corpus for withdrawing a claim: no edge does it, and the only route is an edit to the owning spec's frontmatter. On an `approved` spec that edit is a human's decision, so the plan must carry `acknowledge_approved: true` on that entry or the whole plan is refused. The flag is a human's to write, the way a `Spec-Drift-Waiver` is.

**The exclusions are the hard part.** Eleven of those 81 occurrences had to survive, every one because it is a sentence about the past or about an absence: `AGENTS.md`'s own history line, spec 093's `MUST NOT exist`, the absence assertion `! test -e .claude/rules`, and the scaffold's negative assertions. A tool that rewrote them would have produced a corpus that compiles, passes every gate, and lies in five places. Three clauses answer that (negation, historical file, historical section), and every skip is reported with the clause that spared it: silence is the failure, since the eleven survivors were found by reading a `grep`, and a tool that keeps them without saying so has only moved the reading.

## What is built

- `compact.rs` gains the retire plan, the six forms reduced to the three that take a replacement (D-2), the unit actions and the skip clauses.
- `crates/spec-spine-core/tests/retire.rs`: 15 tests, each exclusion with the sentence it protects and each refusal with the plan that earns it.
- `cmd_compact.rs` and the CLI help carry it; no new subcommand.

`CompactPlan` gained a field, which broke every struct literal that built one, exactly as `ScaffoldFile` did before it. The literals now construct with `..Default::default()` and the type carries the note, so the next field added breaks nothing.

## Evidence

`spec-spine verify 097` passes (15 commands). `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `cargo fmt --check`, `clippy -D warnings` and the workspace suite are green. The two `extends` edges onto 096's files were added by this build rather than by the filing: `V-017` refuses an `extends` onto a unit its owner has only `planned`, so they could not be declared before 096 was built.

Lifecycle flipped to `implementation: complete`; `status` stays `draft` for a separate ratify.


bartekus:
All three addressed in 6c0f9aaa. The first was a real gap in the build, not a nit.

**`UnitAction` validated and never applied: correct, and the worst kind of defect this spec exists to prevent.** A plan could name a withdrawal on an approved spec, carry the human's `acknowledge_approved: true`, pass validation, and leave the frontmatter claiming a path that was gone. The acceptance did not catch it because it asserted the call returned `Ok`, which it did. You named that too, and it is the sharper half of the finding: permission to make an edit is worth nothing if nothing checks the edit happened.

The edit is made now, line-scoped inside the frontmatter and inside the named edge's list, and a withdrawal that empties a list takes the key with it, because a bare `establishes:` with no items is not the document minus a claim. Three tests read the resulting bytes: one per action kind, one for the emptied key (D-2).

Writing that surfaced a second defect of the same shape: the closing `---` of a frontmatter block starts with `-`, so a check for "does a list item follow this key" read the delimiter as an item and kept every key it was meant to remove. The test that catches it asserts the key is gone rather than that the call succeeded.

**`replace_bare` admitting `_`: correct.** `_` is not alphanumeric and was not in the exclusion set, so `_rules/one.md` was rewritten as a citation of `rules/one.md`. Both `_` and `-` are path characters and are named explicitly now, with a test covering a leading underscore and an embedded hyphen (D-3). The existing test only covered a `/`-prefixed predecessor, which is exactly why it passed.

**The `extends` edge on `cmd_compact.rs`: right that the diff did not evidence it.** The resolution is the one you offered second: the file should have changed. §3.5 requires a skipped occurrence to be reported with the clause that spared it, the library carried `skipped` from the first commit, and the CLI printed none of them. It prints them now, on every run rather than only under `--plan`, since the exclusions are where a retirement goes quietly wrong.

On `spec_statuses` running a full compile: accepted as noted. The verb is human-authored and infrequent, and the compile is skipped entirely when `retire` is empty.

bartekus:
Both bugs were real and are fixed in 9ab06e52. They are the same shape as the last round: a tool that edits what the plan did not name, and a report that describes a file other than the one emitted.

**1. Unbounded `contains` in `apply_unit_actions`: correct.** Retiring `rules/` would edit a frontmatter unit claiming `kit/rules/sub/`. You also named the reason it happened: the prose matcher had a boundary test and the unit matcher did not.

The obvious fix, sharing `replace_bare`'s predicate, is wrong, and trying it is how I learned why. It withdrew every unit in the fixture, because a frontmatter value *is* quoted (`path: "rules/"`) and the prose test disqualifies a quoted match **on purpose**: the backticked citation and the quoted `path` form are both replaced before `replace_bare` runs, so matching them again rewrites one occurrence twice. So there are two tests, not one. They share `is_path_char` and differ in exactly the quote rule, and the difference is stated where each is defined (D-4).

**2. `continue` instead of `break` in `retire_file`: correct, and the sharper half is your last sentence.** Both existing skip tests use a single `RetireEntry`, so the interaction between entries was never exercised. A line is spared now, full stop, and the new test emits with two entries and asserts the emitted file matches what the report claims about it (D-5).

**3. One-line lookahead in `drop_empty_edge_keys`: correct.** A key separated from its first item by a blank line read as empty and would have been dropped along with its claims. It looks past blank lines now, with a test that keeps a surviving sibling claim.

On `spec_statuses`: unchanged, and skipped entirely when `retire` is empty.

That is the second remediation round on this pull request. If a further pass finds something new and real, it goes to a human rather than to a third round.

bartekus:
Both bugs were real and are fixed in d4e1aaa0. **This is a third remediation round, past the two a driven session is given**, so the merge is not mine to take and is left to a human. Recording why the round was taken: both findings are correctness defects in code written in the previous round rather than style, and the pass named new code paths rather than restating a declined note, which is the convergence test I would otherwise stop on.

**1. The `path` form preempting a unit action on the same line: correct, and uncovered.** With the form rewrite first, `path: "rules/"` became the replacement, the unit matcher then found nothing, and the action silently did not fire. A withdrawal left the claim standing; a retarget wrote a path the `to` never named. The frontmatter edit runs first now, and the acceptance carries a plan with both a `path` rule and a withdrawal on the same spec (D-6).

**2. `{:?}` as the contract for the `approved` comparison: correct, and the direction matters.** The failure is silent and permissive: a rename or a variant with data would make every approved spec editable without the human acknowledgement, which is the one thing §3.3 exists to demand. It matches the enum now (D-7). The same call also stopped conflating "not approved" with "not in the corpus" — the two are separate reads and separate refusals.

**3. `Withdraw` with a `to`: agreed, and now a refusal rather than a warning.** The field was discarded silently, and a plan whose author wrote one meant something the tool will not do.

On the two notes not actioned: `approved_specs` still compiles the corpus, and the heading match is still a substring test. Both are accepted as noted — the first because the verb is rare and the compile is skipped entirely when `retire` is empty, the second because a heading is authored prose and an exact-equality rule would break on a trailing period. Neither is a defect; if either should change, it is a decision rather than a fix.

Everything else is green: `verify 097` passes 15 commands, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` against the default branch, `fmt`, `clippy -D warnings`, and 23 + 26 tests in the two suites.

bartekus:
Round four, in 22c1926f. Two actionable, one answered with a test.

**B-1, the period: correct, and the inconsistency was internal.** `is_path_char` counts `.` as part of a path, and three lines below it the after-terminator list carried a bare `.`, so `rules/one.md.bak` satisfied both ends of the boundary and was rewritten as a citation of `rules/one.md`. A period terminates a citation only when nothing but space follows it (D-8), with a test carrying both spellings in one line.

**B-2, the double compile: correct, and it was mine from the round before.** Splitting "is the spec there" from "is it approved" cost two compiles of the whole corpus for one validation. One call answers both now.

**B-3, the `break` and a duplicated unit line: the premise does not hold, and it is now asserted rather than argued.** The `break` ends the search for *this line's* action; the next occurrence is the next iteration of the line loop, so every duplicate is withdrawn. A test writes a frontmatter with the same unit twice plus a sibling claim and asserts both copies go and the sibling stays. That is a better answer than a code reading, and the finding is what prompted writing it.

On the traversal note: agreed with your own reading. `repo_root.join(&e.path)` feeds an `exists()` check and then a scan bounded by `scannable_files`, which walks from the repository root with the derived, state and excluded trees already pruned, so a `../` path can only fail the existence check.

Everything green: `verify 097` passes 15 commands, `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` against the default branch, `fmt`, `clippy -D warnings`, 25 + 26 tests.

bartekus:
Round five, in 0bb19ee7. All three were real, and the first is the largest finding on this pull request so far.

**1. Absent form keys and the stated contract: correct, and the resolution is that §3.7 was never built.** You read the doc comment's stronger claim and found the code did not implement it. What neither of us should have needed to infer is that the spec *also* states it, in §3.7: "Exit 1 after the rewrite, with the list: any occurrence of a retired path that is neither rewritten nor skipped by a named clause." Nothing implemented that. So a form no rule covered left the path in place in silence: the plan looked applied, the gate was green, and the corpus still named a path that was gone.

It is built now, and the scan reads every file the rewrite **read** rather than the files it emitted. That distinction was the bug in my first attempt at it: a file no rule touched is exactly where an unaccounted occurrence hides, and it is not in the emitted set at all, so the first version of the test passed against an empty list (D-9). The CLI reports the list and exits 1.

**2. A spared line hiding a second entry's occurrence: correct, and better fixed than acknowledged.** You suggested writing the tension into the spec. It is cheaper to make the report true: "reported, never silent" is a claim about occurrences, not lines, so every occurrence on a spared line now gets its own record naming the clause that spared it. The test counts two records on the one line rather than asserting the emitted text (D-10). You were also right that my previous test accepted the silent survival as correct output, which is why it passed.

**3. `replace_glob` garbling a non-`None` replacement: correct.** A glob rule substitutes the path prefix and leaves the wildcard, so `rules/*.md` with `AGENTS.md` reads `AGENTS.md*.md`. A replacement that is not a directory prefix is refused, naming what it would have produced; `~` still removes the pattern outright (D-11).

**Minors:** `\r` joins the terminator set. The CRLF-to-LF normalisation on a retarget is left as is and is now the only unremarked normalisation in the pipeline, which is consistent with how the rest of the rewrite treats line endings; if that should change it is a decision for the whole verb rather than for one branch of it.

On `corpus_and_approved` running after `read_corpus`: noted and unchanged. Both compiles are the same call on the same tree and the second is skipped entirely when `retire` is empty; collapsing them means threading the outcome through `read_corpus`'s contract, which is a refactor rather than a fix.

bartekus:
Round six, in 6b826a02. All three actioned.

**1. `unaccounted` false positives on a longer path: correct, and the sharpest kind of finding.** Two readers of one grammar have to be the same reader, and they were not: the rewrite used §3.2's boundary rule, the leftover scan used a raw substring test, so `kit/rules/one.md` was correctly left alone and then reported as an occurrence no clause accounted for, refusing the whole run over a path that was never retired. You also named why the existing test missed it: `a_longer_path_is_not_a_retired_one` asserted the rewrite left it alone and said nothing about the leftover list. It asserts both now (D-12), which is the assertion that would have caught this the moment §3.7 was built.

**2. Clause attribution on a multi-entry spared line: correct.** Labelling entry B's occurrence with entry A's clause is accurate about the outcome and wrong about the reason. Each entry's own clause is decided first now, with the line-level clause as the fallback for an occurrence that has none of its own (D-13). The test spares a line by `historical_files` for one entry and by a negation for the other, and asserts both labels appear.

**3. CRLF on a retargeted line: fixed rather than declined again.** It came up in two passes, and it is cheaper to preserve the ending than to keep explaining the inconsistency. Rewriting a CRLF file's retargeted lines to LF changes bytes the plan never named, in exactly the files it did name.

`verify 097` passes 15 commands; `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`, `couple` against the default branch, `fmt`, `clippy -D warnings`, and 29 + 26 tests.

bartekus:
Round seven, in d301c0c3. All three actioned.

**1. `replace_bare` infinite loop on an empty path: correct, including the reason the guard was missing.** `repo_root.join("")` is the repository root, so the existence check passed it. Refused now, with a test (D-14). Either answer the alternative gives is wrong: §3.7's refusal list becomes the whole tree, or the scan never terminates.

**2. Raw `contains` in the clause computation: correct, and this is the same defect as last round one function over.** What I fixed then was the leftover scan; the clause computation was still reading a different grammar. Your trace goes further than the symptom and is the part worth recording: a spurious skip record does not merely add noise, it makes §3.7 treat the whole line as accounted for, so a real unaccounted occurrence beside it goes unreported. A test now asserts a negation on a longer path produces neither a skip record nor a leftover.

**3. The second unit action on a line: fixed rather than documented.** Every matching action fires, with a withdrawal winning over a retarget on the same line, because the line is leaving and there is nothing left to point elsewhere (D-15). The `paths:` sugar is the shape that actually writes two units on one line, so the test uses it.

Two false starts writing that test, both caught by the suite rather than by me: an inline YAML sequence is not a valid unit form in this grammar, so the fixture failed corpus validation instead of asserting anything, and the edit that fixed it changed the edge name in a different test matching the same pattern.

`verify 097` passes 15 commands; `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `couple` against the default branch, `fmt`, `clippy -D warnings`, and 32 + 26 tests.

bartekus:
Round eight, in b93c5aa0. Both correct, and together they name a pattern I should have acted on two rounds ago.

**1. No right-boundary check in the leftover scan: correct.** `rules/one.md.bak` was correctly left alone by the rewrite and then reported as an unaccounted occurrence, refusing a legitimate corpus. You also found the test gap that hid it, and it is the same gap as last round: `a_period_terminates_a_citation_only_at_the_end_of_a_sentence` asserted the emitted bytes and said nothing about the leftover list. It asserts both now.

**The pattern.** This is the third consecutive pass finding the same defect one reader along. Round six: the leftover scan used a raw `contains` where the rewrite used the boundary rule. Round seven: the clause computation did. Round eight: the leftover scan had the left boundary and not the right. Each time I fixed the reader you named, which is how a bug survives three fixes.

So this round does not patch a fourth copy. There is one `occurs_as_path`, taking the context that decides its delimiters, and both boundaries are tested in both contexts. In prose a quote means another form owns the occurrence (the backticked and quoted forms are replaced first); in a quoted value a quote IS the delimiter. The rewrite, the clause computation, the unit matcher and §3.7 all read it, and `preceded_by_path_char` no longer exists to be copied. §3.7 asks whether *either* context would have matched, because a scan narrower than the rules refuses a corpus the rules were right to leave alone (D-16).

**2. Per-entry skip matching: correct, and now structural.** `Skipped` carries the retired path it is about, so a record holding only the coordinate can no longer let one path's skip account for another path's occurrence on the same line (D-17). It also makes the CLI's left-alone report say which path each line was spared for.

`verify 097` passes 15 commands; `check --fail-on-unresolved --fail-on-warn`, `couple` against the default branch, `fmt`, `clippy -D warnings`, and 32 + 26 tests.

bartekus:
Round nine, in b5edb3f5. All four actioned.

**1. The glob form was not boundary-aware: correct, and it is the reader D-16 missed.** Last round converged four readers on one rule and left a fifth on a raw substring test, so `rules/*` matched inside `extra-rules/*.md` — and because the glob branch returns early, the citation rule that would have handled the line correctly was never reached. That early return is what makes this worse than a stray rewrite. The glob is located with `occurs_as_path` now, and `replace_glob` rewrites the occurrence that was found rather than every substring match (D-18).

**2. The `line_clause` fallback: stated in the spec, as you suggested.** D-5 now carries its own consequence rather than leaving it to a reader of the tests: a negation is a line-level marker, so a second path on the line is spared and labelled `negation` even though the `!` does not negate it. Sparing is per line because the report must describe the emitted file and the emitted line is whole; the clause is the best available label, not a claim about which marker applies to which path.

**3. The report omitted `path`: correct, and it is my own gap from one round earlier.** I added the field precisely for the two-paths-on-one-line case and then did not print it, so two records differed only in what the report left out. It prints `rel:line [path clause] text` now.

**4. Doubled doc comment: correct.** An artifact of one of my own splices.

`check` is fresh, `couple` clean against the default branch, `clippy -D warnings` and `fmt` clean, 33 + 26 tests.

bartekus:
Round ten, in c4dc694b. All three actioned.

**1. Raw `String::replace` in the retarget: correct, and it is the sixth reader.** D-16 converged four, D-18 found a fifth, this is the sixth. With `rules/` and `rules/one.md` on one frontmatter line, retargeting the directory rewrote the sibling too and produced a path the plan named nowhere. Every *replacement* of a retired path now goes through one function taking the context, the way every *test* of one already did (D-19).

Six rounds of the same finding is worth naming plainly: it is an argument for the shared function, not for six individually careless readers. Each time I added a reader I re-derived the rule instead of calling it, and each fix converged only the readers under discussion.

**2. `e.path` escaping the corpus: correct.** `repo_root.join("/etc/passwd")` is `/etc/passwd` and `join("../x")` leaves the tree, so the existence check passed both and gave exactly the false confidence you describe. Nothing outside the corpus is ever written, because the walk is corpus-bounded, but the string would then be matched across every scanned file. Absolute paths and `..` components are refused, with a test covering both spellings (D-20).

**3. Empty `to`: correct.** Refused alongside the absent one; both write `path: ""`, a unit no corpus can resolve. The guard reads `trim().is_empty()`, so whitespace does not slip through either.

On the linear scan in `unaccounted`: accepted as noted. The verb is a rare human-authored operation and the lists are per-run rather than per-corpus, so an index would buy nothing measurable here.

`check` fresh, `couple` clean, `clippy -D warnings` and `fmt` clean, 36 + 26 tests.

bartekus:
Round eleven, in 3d22d69b. All three actioned, including the one you flagged as a note.

**1. Directory kind without a trailing slash: correct, and the consequence is worse than a missed rewrite.** Without the slash the path is a *word*, so `rules` matches the English word in any sentence, §3.7 reports every such line as an unaccounted occurrence, and the run is refused over prose. Refused at validation now (D-21). Your framing is the part I want to record: every plan written so far spelled the slash by convention, and convention is not a rule.

**2. The synthetic-occurrence misattribution: taken rather than noted.** You were right that it needs two entries whose replacement text and retired path overlap, which no plan here would write. But it has a principled answer that costs one line: an entry acts on occurrences the **source** line arrived with, never on one another entry's replacement created. Acting on such an occurrence rewrites something no plan named; sparing it records a clause that was never true of the file (D-22). A test now builds exactly that overlap and asserts the synthetic occurrence is neither rewritten nor reported.

**3. `glob_at` building the same pattern twice: correct.** For a directory path `"rules/" + "*"` and `"rules" + "/*"` are one string. Deduplicated.

`check` fresh, `couple` clean against the default branch, `clippy -D warnings` and `fmt` clean, 38 + 26 tests.

bartekus:
Round twelve, in 513196d5. Four actioned; one was already fixed.

**3 (the markdown link label) is the sharpest finding on this pull request, and it is a different shape from the eleven before it.** Every earlier round of this family was two readers disagreeing about whether something matched. This one is §3.7's scan being **broader than the rules it scans for**: `[rules/one.md]` passes the value context because brackets are not path characters, no form rewrites a link label, and the run therefore refuses a corpus that no rule could have repaired. Marking the file historical would have been the operator "fixing" a tool that was wrong. A label is a citation, so the prose context delimits on brackets and the citation form covers it (D-23).

**2 (the `path` form's raw `str::replace`): correct, and it was the last one.** You are right that quotes bind tightly enough to make it safe in well-formed YAML, and right that the D-19 rule was not applied consistently. It goes through `replace_path_in` now, with a test whose replacement text contains the retired path so the double-fire you describe would be caught.

**The `./` edge note: taken as a refusal.** It is worse than surprising: the run rewrites nothing, and then §3.7 reports every bare occurrence as unaccounted for, so the operator gets a refusal listing the whole corpus for a typo in the plan. Refused at validation, naming the prefix (D-24).

**The performance note: taken.** The compile answers a question only the unit actions ask, so it now runs only when some entry names units.

**1 (duplicate glob patterns) was already fixed** in 3d22d69b, the round before this pass ran; that review saw an older head.

`check` fresh, `couple` clean, `clippy -D warnings` and `fmt` clean, 41 + 26 tests.

bartekus:
Round thirteen, in 258cf8c5. The line-number finding was real and precisely traced; thank you for the worked example, which is what made it quick to reproduce.

**The coordinate mismatch: correct.** A glob deletion removes a line from the output, so every later skip record's source line number runs ahead of that line's position in the emitted file, and §3.7 reported a correctly spared occurrence as unaccounted for.

Of the two fixes you offered I took neither exactly. Renumbering the records against the output would make the report useless to a reader of the original file, and threading the rewrite list into the scan does not distinguish a line that was rewritten *for one path* from the same line still carrying another. Spares are matched on the line's **text**: a spared line is emitted unchanged, so its text identifies it in both, and the report keeps source numbers (D-25).

Two notes on the test, since the first version of it was wrong in an instructive way. It asserted the whole leftover list was empty and failed for the *right* reason on a different file: `000-alpha` claims `rules/` in its frontmatter and that plan declares no `path` form or unit action for it, so §3.7 was correct to report it. I scoped the assertion to the file under test rather than weakening it. Then I reverted the fix and re-ran, confirming the test fails without it, so the assertion is known to be able to fail rather than assumed to be.

**The `line_clause` design note:** agreed, and it is D-5 plus D-22. The test you name asserts the own-clause path, which is the one that was wrong before; the fallback remains deliberate and is documented as a label rather than a causal claim.

**Performance:** the compile was narrowed in the previous round and now runs only when some entry names frontmatter units, so a forms-only plan pays nothing.

`check` fresh, `couple` clean, `clippy -D warnings` and `fmt` clean, 42 + 26 tests.

bartekus:
Round fourteen, in e9667745. Two failures on the previous head, and the first was mine rather than yours: CI refused on `cargo fmt --all --check`. I ran the formatter, edited the test again, and committed. Noting it because it is the second time this session a gate caught something I had already "done".

**The synthetic-occurrence leftover: correct, and the assertion gap is exactly where you put it.** D-22 stopped the synthetic occurrence being *spared*; it was still *reported*, so the run would exit 1 on a corpus handled exactly as the plan asked. My test asserted the emitted content and the absence of a skip record and never asserted the leftover list, which is why it stayed green over a refusal.

I did not patch the scan again. §3.7 is now decided inside `retire_file`, per line, with the source line and its result both in hand. From that position neither bug is expressible: it cannot mistake an occurrence the line arrived with for one a replacement text created, and there are no coordinates to shift, so last round's text-matching goes away with the whole-file scan it was working around (D-25; the superseded form is kept as D-25a because its reasoning is still the argument against coordinates).

That also answers the third of your notes structurally: there is no second corpus-level pass over the emitted files at all now.

**The `path` form's quote-delimited replace:** already through `replace_path_in` as of 513196d5, the round before this pass ran. Your reasoning for why it was safe is right, and it is also the reason it went unnoticed for six rounds.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, and 42 + 26 tests all clean locally before pushing this time.

bartekus:
Round fifteen, in 0cfa58dc. Both bugs actioned; both performance notes are already gone.

**The empty `historical_sections` keyword is the most dangerous finding on this pull request.** `"any heading".contains("")` is true, so one empty string spares every occurrence in every file: the retirement does nothing, the report says everything was left alone deliberately, and the run exits 0. Every other refusal in §3.7 exists to stop a silent *partial* rewrite; this one stops a silent *total* one, which no assertion in the suite would have caught because every file legitimately "has no leftovers". Refused, with empty `historical_files` entries alongside it (D-26).

**The glob early return: fixed rather than documented.** You framed it as a design note, and a deletion genuinely does take the whole line, so nothing else can apply. But a *replacement* leaves the rest of the line standing, and returning there stranded a citation sharing it for §3.7 to refuse — on a corpus the rules could repair (D-27). That is the same shape as D-23: the tool refusing what it was capable of handling. The test puts a glob and a bare citation on one line and asserts both are rewritten and nothing is reported.

**`unaccounted`'s linear scan and `examined`'s second copy of every file are both already gone.** §3.7 moved into `retire_file` in e9667745, the commit before this pass ran, and the whole-file scan went with it: there is no second corpus-level pass and no duplicate buffer. I checked the source rather than assuming, since this is the second pass in a row that reviewed an older head.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, and 44 + 26 tests clean before pushing.

bartekus:
Round sixteen, in f636c5f5.

**`replace_bare`'s relative offset: correct, and you are right that D-16 named exactly this.** A match beginning where the previous one ended takes `occurs_as_path`'s `at == 0` short-circuit and never reads its real left neighbour. The function was `replace_path_in` carrying that bug, so it is **deleted** rather than repaired: D-16 said every reader goes through one function, and a second copy of a function is a reader (D-28). You called it "not through it", which is the accurate description; patching it would have left two again.

**The `own`/`current` divergence: taken.** They were equal at that point, so no live defect, but the comment already claimed the source line and a later reordering would have broken D-22 silently. It reads `line` now, like `present` does.

**Writing the test for this took three attempts, and every earlier version asserted nothing.** Recording them because the pattern is the recurring failure mode in this corpus:

1. `git stash` removed the new test along with the fix, so the name filter matched nothing and cargo reported `0 passed; 44 filtered out` — a vacuous pass I nearly accepted as evidence.
2. The fixture put a separator between the two occurrences, so the walk never reached offset zero of the remaining slice.
3. The fixture then had *only* invalid occurrences, so `names_path` rejected the line before the replacer ran. The test passed identically with and without the bug, and an `out.is_empty() ||` clause hid it. Tightening the assertion to print the real output is what exposed it.

It now fails against the previous commit's source (output shows the abutting occurrence rewritten) and passes against this one. Both directions confirmed rather than assumed.

**`unaccounted` and `examined` are gone**, removed in e9667745 when §3.7 moved into `retire_file`. That is the third consecutive pass reviewing an older head, so I check the source before answering those.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 45 + 26 tests.

bartekus:
Round seventeen, in 2389423a. One finding was new, and it is more serious than the pass judged it.

**The edge-key heuristic: correct, and not benign.** You reasoned it was harmless from `id:` and `title:`, where the next line is another column-zero key so nothing is captured. That is right about those keys and wrong about the shape. A block scalar's continuation lines **are** indented and cannot be told from list items by shape alone, so `summary: >` set the current edge to `summary` and a unit action naming `summary` **deleted a sentence out of an approved spec's frontmatter**. The failing output shows the line gone.

Worth recording that my first test for this used `id:` and passed against the defect for exactly your reason. I only reached the bug after building the block-scalar fixture. It now fails against the previous commit and passes against this one, both confirmed.

Your suggested rule is what landed: an edge key is a line that is exactly `key:`. I added the other half, since it was implied but not stated — any other column-zero key **ends** the list it followed, rather than leaving the previous edge current.

**`replace_bare` and the empty `historical_sections` entry are both already fixed**, in f636c5f5 and 0cfa58dc respectively, the two commits before this pass ran. That is the fourth consecutive review of an older head; I verify each such finding in the source before answering rather than re-fixing it.

**The §3.6 prose note:** read it again as you suggested. "They were added by this build rather than by the filing" is consistent with the frontmatter and with V-017's ordering; no contradiction, and the frontmatter is authoritative either way.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 46 + 26 tests.

bartekus:
Round eighteen, in 5c7877a8. Two findings were real and reachable; the third is not, and I want to be exact about which is which.

**1. The phantom rewrite record: correct, and reachable with a legal plan, as you said.** Two entries can contribute a retarget and a withdrawal naming the same line. The retarget was recorded, the withdrawal then dropped the line, and the report carried a record for a line that never reached the output while §3.6's per-form count included it. The withdrawal is looked for first now and the record follows the decision (D-30). The test fails against the previous commit.

**3. `historical_files` traversal: correct, and I read it the way you framed it.** No security consequence, exactly as you say, because the comparison is literal against corpus-relative paths. The defect is silence: a mistyped `../outside.md` excludes nothing, reports nothing, and the retirement proceeds without the exclusion its author intended. Refused with the same rule the retired path carries (D-31). Also fails against the previous commit.

**2. The `ends_with(':')` predicate is not reachable in a corpus that compiles.** I tried to build the failing case and could not. A quoted scalar ends with `"`, not `:`, so `summary: "See rule:"` never matches the predicate; and an unquoted scalar ending in a colon fails to parse, which `compile` refuses before `compact` runs. I verified that rather than reasoning about it: the fixture returns `Config("the corpus does not validate")`.

I fixed it anyway, because "an undeclared assumption on the call site" is a fair description and the scan should use the walk's key rule. But the test asserts the rule rather than reproducing a defect, and its doc comment now says so. A test that cannot fail should at least admit that it cannot, rather than sit in the suite looking like coverage (D-32).

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 49 + 26 tests.

bartekus:
Round nineteen, in b7bec83f.

**The within-call re-read: real as a structural weakness, no reproduction found, fixed anyway.** You framed it precisely — "the tool should not rewrite what the form replacement created", and it is D-22 inside a single `apply_forms`. The forms now collect spans against the **source** line and substitute once, earliest first, dropping overlaps, which is the rule spec 096 §3.4 already applies to ids.

I tried to build the failing case and could not. Every replacement text that embeds the retired path with a clean *left* boundary fails the *right* one, and the reverse. So the sequential form was safe today by a coincidence of the two boundary rules rather than by construction, and a property that holds by coincidence is one nobody can rely on while editing either rule. The test's doc comment says it asserts the rule rather than reproducing a defect.

**The `line_clause` fallback: you were right that no test covered it.** `each_occurrence_on_a_spared_line_carries_its_own_clause` only exercises the case where every entry has its own clause. The documented fallback is now asserted directly, so it is design rather than accident.

**The phantom rewrite record was fixed in 5c7877a8**, the commit before this pass ran. That is the fifth consecutive review of an older head.

One note on process: the span rewrite left `replace_glob` unreachable, and `clippy -D warnings` caught it as dead code **before** the push. Two rounds ago CI caught a formatting slip after a push. Running the full gate before pushing rather than after is the difference between those two rounds.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 51 + 26 tests.

bartekus:
Round twenty, in 307a91b1. Two real defects, both with tests that fail against the previous commit.

**1. The glob replacing only the first occurrence: correct, and worth the check.** You named it against `replace_glob`, which the previous commit deleted, so the easy answer was "already gone". It was not: the span form that replaced it pushed one glob span, so an array carrying `"rules/*.md", "rules/*.toml"` had the second survive and §3.7 refuse a corpus the rules could repair. The finding was about the *answer* — every other reader returns every occurrence and the glob returned the first — not about the function it lived in (D-34).

**2. The sub-heading losing its parent: correct.** `historical_sections: ["History"]` stopped matching the moment a `###` appeared under `## History`, and every occurrence below it was refused. The tracker keeps the heading stack now and a keyword matches any ancestor. The test also asserts the section **ends** at the next sibling, since "matches any ancestor" could otherwise spare the rest of the file.

**3. The block-scalar edge is already fixed**, in 2389423a. That round found it deletes prose from an approved spec's frontmatter rather than being latent, and the rule you propose here — restrict `edge` to structural keys — is close to what landed: an edge key must be a line that is exactly `key:`, and any other column-zero key ends the list.

**4. The double `names_path`: correct.** `own` reads the `present` it already computed.

Six passes in a row have now reviewed an older head. I check each finding against current source before answering, which is what turned finding 1 from "already fixed" into a real bug.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 53 + 26 tests.

bartekus:
Round twenty-one, in d55aa367.

**The backtick form: correct, and the fix nearly did more damage than the gap.** You were right that it was the last raw substring match and right about why it was safe — "only because backticks act as de-facto boundary markers", which is the same "safe by coincidence" reasoning D-33 was written against.

My first attempt applied `is_path_char` to both sides, which is what the other readers use. That refused **every citation ending a sentence**, because `.` is a path character for extensions. The suite caught it on the first run. The backticks *are* the delimiters, so the test is that neither side abuts another backtick (we would be inside a longer code span) or an identifier character (the backtick belongs to a longer token). Sentence-ending and comma-following citations are pinned by a test now (D-36).

Recording that because it is the second time in this review cycle that reusing a rule "for consistency" was wrong: the two path contexts needed different quote handling (D-4), and a delimiter is not a boundary character.

**The negation reading `current`: correct.** It was the one clause D-22 left reading the mutable buffer. Equal to `line` there today and not after any reordering, which is precisely the failure you describe. It reads `line` now.

**`own[i].is_none()` being redundant: correct, and kept deliberately** with a comment saying so. It states what the loop means, and a later `line_clause` that did not cover every entry would otherwise start reporting spared occurrences without any line changing.

**The `historical_files` path checks landed in 5c7877a8** — absolute, `..`, and `./` are all refused there, with a test. Seventh consecutive pass on an older head.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 54 + 26 tests.

bartekus:
Round twenty-two, in 3ca6882f. **This is the first pass with no actionable defect**, and I want to show the work rather than assert it.

**The `glob_at` context: a real observation with the wrong conclusion, and I measured the alternative rather than arguing.** You are right that `PathContext::Value` accepts a backtick on the left, so `` `rules/*.md` `` satisfies `glob_at`. I implemented your proposed fix — prose context — and ran it:

```
OUT="" LEFT=[Leftover { rel_path: "docs/note.md", line: 1,
  text: "Patterns live in `rules/*.md` today.", path: "rules/" }, ...]
```

The line is left unrewritten **and** reported as unaccounted for, so the run exits 1 on a corpus the rules can repair. The citation form cannot pick it up either: the exact backticked path `` `rules/` `` is absent from `` `rules/*.md` ``, and the bare occurrence is disqualified by that same backtick. Your alternative suggestion, running the backtick citation before `glob_at`, has the other problem — a citation replacement is prose, and splicing prose in front of `*.md` yields neither a pattern nor a sentence.

So the current behaviour is right: a backticked glob is a glob, and the glob form replaces its prefix. It is pinned now by a test whose doc comment carries your proposal and the measurement, so the next reader reaching for this change finds the answer instead of the question (D-37).

**The `line_clause` label: documented design.** D-5 records it as a label rather than a causal claim, D-13 gives each occurrence its own clause where it has one, D-22 stops synthetic occurrences reaching it, and `an_entry_with_no_clause_of_its_own_takes_the_lines` has asserted the fallback since the nineteenth pass.

Nothing changed in the implementation this round. `fmt --check`, `clippy -D warnings`, `check`, `couple`, 55 + 26 tests.

bartekus:
Round twenty-three, in 9ba1b630.

**The fenced-block heading finding is correct, and your framing of why it matters here is exactly right.** Every spec in this corpus carries `verify:cli` blocks full of `# comment` lines, so `# Live paths follow` became the current heading: a `historical_sections` keyword naming the real section stopped matching inside the block, and one that happened to match a comment spared every citation after it. Fences are tracked now, with `~~~` as well as backticks.

**My first fix for it was worse than the defect, and I want that on the record.** I skipped fenced lines entirely — `continue` before any rule ran. That would have hidden every occurrence inside a `verify:cli` block from §3.7: silent omission, which is the precise class D-26 exists to refuse, introduced while repairing something else. I caught it reading my own diff rather than from a failing test, which is luck I would rather not rely on. Only the **heading update** is suppressed now; the line is rewritten, spared and accounted for exactly as any other, and the test asserts both halves.

**The negation check already reads `line`**, fixed in d55aa367 with the same reasoning you give: equal to `line` today, and not after any reordering. That is the eighth consecutive pass reviewing an older head, so I check each finding against current source before answering — which is also how the previous round's glob finding turned out to be real rather than stale.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 56 + 26 tests. The new test fails against the previous commit and passes against this one.

bartekus:
Round twenty-four, in a12d2f79. Both findings were real, and the first is the sharpest structural observation on this pull request.

**1. D-22 was half a rule, and you found the other half.** The `present` gate stops an entry acting when its path was absent from the source line. It cannot see the case where the path **was** present and a previous entry's replacement created a second occurrence — the entry then rewrote both, and only one was the plan's subject. You also named why the existing test missed it: `an_occurrence_another_entrys_replacement_created_is_not_spared` only covers `present[B] = false`.

I took the design change you proposed rather than a narrower patch: spans are collected from the source line across **all** entries and applied once. D-33 established that discipline within a single `apply_forms`; this extends it across entries, so the rule is now identical at every level — find every match against the source, then substitute once, earliest first, dropping overlaps.

**2. The `path` form reaching inside a backtick citation: correct.** A `path`-only plan replaced the path inside `` `rules/one.md` `` and left the backticks wrapped around prose. The form is about a quoted value, so it requires the quotes now. The glob deliberately keeps the wider rule, because D-37 established by measurement that a backticked glob is a glob and narrowing it there causes a refusal on a repairable corpus.

Worth naming: that is the fourth time these two contexts have needed to differ in a specific way rather than share a rule — D-4 (quotes), D-36 (delimiters are not boundary characters), D-37 (globs keep the wider rule), and now this. The proliferation is the point rather than a smell: each difference is a real distinction between prose and a quoted value, and collapsing them has failed four times.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 58 + 26 tests. Both new tests fail against the previous commit.

bartekus:
Round twenty-five, in 68deeae4.

**The implicit multi-line scalar: correct, and it is the half D-29 left behind.** I fixed `summary: >` in round seventeen and left `summary:` with an indented sentence under it — still read as a list, so a unit action naming `summary` deleted the sentence out of the frontmatter. An edge key now opens a list only when a list **item** follows, which is the rule `drop_empty_edge_keys` already used at the other end; both ends of the walk finally agree about what a list is (D-41). The test fails against the previous commit.

**The `retired-path` count: your inference was wrong, your point was right.** The increment does exist — `compact.rs:297`, `*counts.get_mut(r.form.as_str()) += 1` — so the counter moves; it was outside the diff hunks, as you suspected it might be. But the coverage gap you identified from that is real and is the more useful half: every test asserted the key **exists**, or that it is zero on an *empty* plan, so the counter could have sat at zero through a live rewrite without a single test failing. There is now a test asserting it is non-zero and equal to the number of records (D-42).

That is the assertion those tests only looked like they were making, and it is the recurring failure mode on this pull request — I would not have found it by reading them.

**`glob_spans` dedup:** sorted and deduped rather than a linear scan per push.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 60 + 26 tests.

bartekus:
Round twenty-six, in 1f72d3d2.

**The contradictory output on a refused run: correct, and now fixed.** The summary printed first and the §3.7 refusal after it, so a reader got two accounts of one run — counts and skipped lines implying work had been computed, then "Nothing was written". The refusal prints first and the summary does not print at all. A report that has to be reconciled with the line below it is worse than no report (D-43).

**The D-22 finding is the previous pass's, and it was fixed in a12d2f79.** Spans are now collected from the source line across every entry and applied once, which is the fix you describe in your first paragraph ("run all entries against `line` … then apply the merged span set once — the same approach `apply_forms` itself uses internally"). The case you work through, where B's path is in the source *and* A's replacement creates another occurrence of it, is covered by `an_entry_does_not_act_on_an_occurrence_another_entry_created`, added that round; it fails against the commit before it.

That is the ninth consecutive pass reviewing an older head. I verify each such finding against current source before answering rather than either re-fixing it or dismissing it — which is what turned the glob finding four rounds ago from stale-looking into a real bug.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 60 + 26 tests.

bartekus:
Round twenty-seven, in d5986e54. The `CurDir` finding is real and **your proposed fix does not work** — I implemented it exactly as written and the test still failed.

`Path::components()` normalises an interior `.` away:

```
rules/./one.md -> ["Normal(\"rules\")", "Normal(\"one.md\")"]
./x.md         -> ["CurDir", "Normal(\"x.md\")"]
```

So a `CurDir` arm catches a *leading* `./`, which D-24 already refused, and can never catch `rules/./one.md`. The test has to be on the literal string, which is what gets compared against the corpus anyway. Had I written the guard without a test that exercises it, I would have shipped a check that looks correct, matches the review's advice, and does nothing (D-44).

The defect you identified is real and is the important half: `rules/./one.md` resolves, the existence check succeeds, the literal string then matches nothing, and the run rewrites nothing, reports nothing and exits 0. That is D-26's silent total no-op arriving through the path field. Refused now, with the same rule applied to `historical_files`.

**The `--plan` message: decided, as you asked.** A refused dry run now says the plan is not ready to apply. "Nothing was written" under `--plan` answers a question nobody asked and reads as a refusal of the command rather than of the plan (D-45).

Two of my own edits this round were caught by the suite rather than by me, and I would rather record them than not: a `replace(..., 1)` matched an identically shaped loop in a *different* test, putting the new case where no guard covered it; and a message replacement silently did nothing on a whitespace mismatch, so an assertion expected wording the code never emitted. A replace without an assertion is a no-op that looks like an edit.

**The performance note stands as recorded**: the compile runs only when a plan names frontmatter units, so a forms-only retirement pays nothing.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 60 + 26 tests.

bartekus:
Round twenty-eight, in d74e666f. Neither finding was a live defect, and both were worth taking for the reason you gave each.

**1. The edge-clearing comment overclaimed: correct.** It said any column-zero line ends the list; the code required a colon. You are right that valid YAML does not produce the gap, and right about why it still matters — the invariant a later reader relies on was not the one enforced. A column-zero line **with content** now ends the list, and a blank line deliberately does not, since YAML allows one between items (D-46).

**2. Two scanners entangled by coincidence: correct, and the last clause is the actionable one.** "A future change to either boundary rule could silently re-introduce double-rewriting **without a test failing**" — that part was true, so the test now exists. It asserts the exact emitted line for a line carrying both a backticked and a bare occurrence, and that one line produces one rewrite record rather than two (D-47).

That is the same reasoning as D-33 and D-36: a property that holds because two rules happen to agree is one nobody can rely on while editing either of them. Three of the defects on this pull request were exactly that shape, so the pin is cheap insurance rather than ceremony.

`fmt --check`, `clippy -D warnings`, `check`, `couple`, 61 + 26 tests.

PR #278, opened 2026-09-21
fix(092): untrack the .agents and .codex trees spec 092 removed

bartekus:
Fifteen files that spec 092 deleted are tracked again on the default branch, and **spec 092's acceptance has been red there since #277 merged**.

## What happened

`.agents/` (10 skill files) and `.codex/` (5 agent and hook files) were untracked leftovers in my working directory. One of my `git add -A` calls swept them onto the spec 097 branch, where they rode through twenty-eight review rounds and merged in `954fb68b`.

This is the second time in this session that `git add -A` committed work nobody asked for. The first was a branch switch carrying a different spec's build into the wrong pull request, backed out in `c16b2078`.

## Why no check caught it

CI's `self-governance` job runs `check`, `lint` and `couple`. It does not run `verify`, which is deliberately outside the gate: `verify` executes what the corpus declares, and the gate runs against branches whose contents are, in the general case, a stranger's.

So **a merged spec's acceptance can go red with every required check green**. That is worth knowing separately from this mistake, and it is how a tree that spec 092 explicitly asserts does not exist came back without a single red light.

## The fix

Both trees are removed from git and from disk. `spec-spine verify 092` passes again (76 commands), including the two assertions that were failing:

```
! test -e .agents
! test -e .codex
```

`spec-spine.toml` needed no change: its only mention of `.agents` is a historical comment, and the hashed input is `.claude/agents/*.md`.

`check` is fresh, `couple` is clean against the default branch.


bartekus:
Both findings are refuted by the corpus, and I checked each rather than asserting it.

**1. "Unwaived coupling-gate violation: spec 093 establishes these files."** Spec 093 establishes exactly six units, and neither tree is among them:

```
- "AGENTS.md"
- { kind: directory, path: ".claude/skills/" }
- { kind: directory, path: ".claude/agents/" }
- ".claude/settings.json"
- "crates/spec-spine-core/tests/harness_hooks.rs"
- "crates/spec-spine-core/tests/harness_skills.rs"
```

`spec-spine index owner` agrees for both deleted paths:

```
.codex/hooks.json               (no spec owns this path)
.agents/skills/build/SKILL.md   (no spec owns this path)
```

And the gate itself answers the question directly: `spec-spine couple --base origin/main` exits 0, and CI's `self-governance` job, which runs `couple`, passes on this PR. A waiver is not missing; there is nothing to waive.

The inline `# Spec 093 …` comments you found are real, and they are why the files looked owned. They were *copied* into a generated projection of `.claude/settings.json`. A comment citing a spec is not a claim by that spec.

**2. "Governance regression risk: four enforcement layers removed."** You asked the right question — whether `.claude/settings.json` carries equivalents — and said it could not be verified from the diff. It can be verified from the tree, and it does:

```
SessionStart: ['session']          PostToolUse: ['recompile']
PreToolUse:   ['pr-gate', 'push-gate']   Stop: ['freshness']
```

All four layers you enumerate are present there, and they are the originals: `.codex/hooks.json` was a generated projection of that file, removed by **spec 092 §3.5** along with `kit/` and `.agents/`.

**The decisive point:** spec 092's acceptance *requires* these trees to be absent —

```
! test -e .agents
! test -e .codex
```

— and that acceptance has been **red on the default branch** since #277 merged, because I swept these untracked leftovers in with `git add -A`. This PR restores the state the corpus demands; `verify 092` passes again at 76 commands. Deleting them is not the drift, their presence was.

PR #280, opened 2026-09-21
feat(098): a citation the renumber could not see

bartekus:
Spec 096 recorded seven defects from spec 095's hand renumber and mechanized the verb that would have prevented them. Two of the seven, the wrapped citation and the multi-ordinal citation, were measured after 096 was filed and never repaired: `compact` refuses a plan naming a spec the corpus no longer has, so it prevents the next renumber, not the last one.

## The measurement

Re-measured against the tree as it stood before the collapse, the debt is **630 occurrences in 146 files**, not the 41 previously counted, and two of its three classes nobody had looked for:

| Class | What | Count | Files |
|---|---|---|---|
| A | a spec document's own title heading names another spec's ordinal | 90 | 90 |
| B | a prose citation (096 §3.3 form 2), wrapped or multi-ordinal | 57 | 37 |
| C/D | a bare ordinal: `084 §3.1`, `050 3.6`, `092 D-3`, `056's`, `086-...` | 483 | 79 |

Class A is the loudest: 90 of 98 documents open with a line naming a spec that is not them. `specs/006-distribution/spec.md` was titled `# 007: Distribution, uvx/PyPI wheel shim`.

## What is built

All 630 are repaired, and the gaps behind them are closed so the next compaction cannot recreate them:

- `compact` gains **form 4**, the document's own title heading, located after the frontmatter and rewritten only when the ordinal is the document's own.
- `compact` gains **form 5**, a bare ordinal followed by a section reference, a decision reference, a possessive or an elided slug, with every exclusion of form 2 unchanged.
- `scannable_files` decides an extension-less file **by its bytes** rather than by a three-name allowlist. `.gitignore` read as an empty stem with the extension `gitignore` and was on no list, which is why eight plain citations survived in it and `.gitattributes`.
- `lint` gains **`L-013`**, warning tier: a title heading that names another spec's ordinal. It is the only one of the three classes a lint can decide, because the directory already states the answer.

`L-013` reported exactly **90** violations on the parent commit, the class-A figure arrived at independently, and reports 0 here.

## Two things the measurement had to learn

- **`git blame` cannot see inside a squash merge** (D-8). The collapse commit is not an ancestor of `main`, so every line the collapse's own PR wrote reads as post-collapse; the first pass was short by five, including `ci.yml` and two in `.gitignore`. The test that holds compares a line's content against the pre-collapse tree. Four of the five are repaired; the fifth narrates the history of a removed spec and rewriting it would make the sentence false rather than stale.
- **The repair changes no acceptance command.** 35 of the 630 fall inside a `verify:cli` block and all 35 are comment lines.

## Evidence

- `cargo test --workspace --locked`: **948 passed / 0 failed**; clippy `-D warnings` and `cargo fmt --all --check` clean.
- `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`: green.
- `couple`: 139 paths checked, no drift, **with no ownership edge added** (D-7 records why that clearance is incidental and was not papered over with `extends` edges).
- `spec-spine verify 098`: passed (24 commands).

## What is left

Roughly 930 three-digit tokens on pre-collapse lines still name a moved ordinal and are left, measured, in §4. Most are not citations at all (`V-001/005/006`, `exit 101`, `feat(017)` as an example commit message, a fixture named `"016-short"`); the genuine ones among them are written in forms a pattern cannot separate from those (`before 084`, `pre-096`, a bare `(101,`).

**Adopter note:** `L-013` is a new warning-tier code, so it is a new refusal wherever `lint --fail-on-warn` is set. The scaffolded template writes a literal `# NNN: Title`, so a corpus that has only ever been scaffolded stays silent.


bartekus:
Pushed a second commit after review: two further modes the first pass's comparison could not see.

- **A reflowed line.** A content comparison fails on a line that was rewrapped since the collapse even when its sentence is untouched; five citations in `docs/design/05-remaining-waves-2026-09.md` survived a prose rewrap still naming a moved ordinal. The measurement now normalizes line breaks and comment markers away and compares the surrounding sentence.
- **A citation carried by hand into a rewritten sentence.** No test finds these. `standards/spec/contract.md` and `standards/spec/constitution.md` each say `spec 040` in a paragraph spec 092's PR rewrote, and 040 today is a different document. Both are repaired because both were read and both unambiguously mean spec 037. The other 155 candidates are not: the only test that finds them is the weak one D-1 rejects, and most of what it returns are correct citations.

637 occurrences in 148 files, up from 630 in 146. D-9 records all three tests and what each missed. Gate, suite (948 passed / 0 failed) and `verify 098` (24 commands) green; `couple` 141 paths, no drift.

PR #282, opened 2026-09-21
feat(099): a merged acceptance is asked again

bartekus:
`verify` executes what a spec declares, so it sits outside the gate chain on purpose: a pull request is, in the general case, a stranger's code. Spec 089 built the sweep that re-asks the corpus's acceptance and made it maintainer-invoked — which leaves the window it was built to close open between runs.

## Measured twice on this repository

- **Spec 092's block** was red on `main` from #277 until #278, with every required check green the whole time. Found by a human reading a diff.
- **Spec 089's own block** was red for longer, and nobody read that diff. It runs `verify-sweep.sh --only 012` and asserts the answer is `011-index-hash-slices`; the collapse renumbered the full id and left the bare ordinal. Found by running the sweep by hand while drafting this spec, and repaired in spec 098 §3.7 (#281).
- A full sweep of `main` at `1c133c20` reached every spec: **54 passed, 43 exempt, 1 failed** — and the one failure is 089's own. The corpus spent **17 minutes** inside `verify`, which is the measured cost of the nightly leg.

## What this builds

- **`.github/workflows/acceptance.yml`** — `push` to the default branch, a nightly `schedule`, and `workflow_dispatch`. No `pull_request` leg, and no `needs:` entry in `ci.yml`, so it sits outside `ci-gate` **by construction** and can never block a merge. That is a structural property, not a promise.
- **`scripts/acceptance-scope.sh`** — maps a revision range to the specs that own a changed path, so the push leg is proportional to the change. An empty scope is a pass with nothing run: a sweep of nothing reporting "all passed" is a green light for a question nobody asked.
- **The nightly leg sweeps the whole corpus** — the only leg that catches a break in a spec that owns none of the changed paths. Spec 092's block asserts an *absence*, and an absence is owned by nobody.
- **`permissions: contents: read` and no secret.** Spec 089 §4 asked for this decision explicitly; §3.4 is it, stated as a requirement so a later `secrets: inherit` is a change to a spec rather than a line in a YAML file.

It amends spec 089's "never CI" to "never on untrusted input" — the rule that clause was written to hold — and corrects one line of 089's acceptance **in place** rather than replacing its sixty commands under `amends_verification`.

## The spec is its own witness

Adding the workflow turned spec 089's block red at command 60, and nothing in the gate chain would have noticed. The change that closes "a merged acceptance can go red with every check green" would itself have gone red with every check green.

## Four defects in the drafting, recorded rather than quietly fixed

| | |
|---|---|
| **D-4** | the run block ended with `echo "sweep exit: $?"`, which makes every sweep green, because a step's status is its last command's |
| **D-5** | three assertions grepped the workflow for `secrets:` and read its own comment explaining why it has none. Both copies now skip comment lines, and were **mutation-tested** against a deliberately broken workflow (`pull-requests: write`, `secrets.NPM_TOKEN`, `git push origin HEAD`): two of five tests went red and both acceptance lines returned 1 |
| **D-6** | the fixture wrote each computed scope *into the repo it was measuring*; the next commit carried it and the scope after that refused on a stale ledger. The refusal was correct and is now asserted deliberately |
| **D-7** | the sweep was corrupted by editing the script it was running from — `bash` reads incrementally. A second argument for CI as the home: a runner's checkout is one nobody is editing |

## Evidence

- `spec-spine verify 099`: passed (54 commands); `verify 089`: passed (68 commands).
- `cargo test --workspace --locked`: 955 passed / 0 failed; clippy `-D warnings` and `cargo fmt --all --check` clean.
- `check --fail-on-unresolved --fail-on-warn`, `lint --fail-on-warn`, `index coverage --fail-on-untraced`: green.
- `couple`: 6 paths checked, no drift.


bartekus:
Pushed one more commit: the workflow arrived pinning `actions/upload-artifact@v4` while `ci.yml`, `determinism.yml` and `release.yml` all pin `@v7`. Now aligned.

Worth noting as a gap rather than just a slip: nothing in the suite catches a workflow that pins a different major from its neighbours. `tests/gate.rs` asserts plenty about `ci.yml`'s *shape*, but not that the repository's workflows agree on the actions they call. That is spec 094's territory rather than this spec's, so it is flagged here and not built.

